PR #13, opened 2023-08-01
Fix remote generation usage conflicts

buildbreaker:
Locally generating and packaging protovalidate protos in this repo.

Fixes: https://github.com/bufbuild/protovalidate/issues/62


gradle/libs.versions.toml
```diff
@@ -17,8 +17,6 @@ mail = { module = "javax.mail:mail", version = "1.4.7"}
 nullaway = { module = "com.uber.nullaway:nullaway", version = "0.10.11" }
 protobuf-java = { module = "com.google.protobuf:protobuf-java", version.ref = "protobuf" }
 protobuf-java-util = { module = "com.google.protobuf:protobuf-java-util", version.ref = "protobuf" }
-protovalidate = { module = "build.buf.gen:bufbuild_protovalidate_protocolbuffers_java", version = "23.4.0.1.20230721020619.4464c06fa399" }
```
buildbreaker, line 20: Primary change

.github/workflows/ci.yaml
```diff
@@ -24,6 +24,13 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
           cache: 'gradle'
+      - uses: actions/checkout@v3
+      - uses: bufbuild/buf-setup-action@v1.24.0
```
pkwarren, line 28:
```suggestion
      - uses: bufbuild/buf-setup-action@v1
```

We can just keep this updated with the latest.

.github/workflows/ci.yaml
```diff
@@ -24,6 +24,13 @@ jobs:
...
+        with:
+          github_token: ${{ github.token }}
+      - uses: arduino/setup-protoc@v1
```
pkwarren, line 31: What is this for?

Makefile
```diff
@@ -60,6 +60,9 @@ help:  ## Describe useful make targets
 
 .PHONY: generate
 generate: generate-license  ## Regenerate code and license headers
+	buf generate --template buf.gen.yaml buf.build/bufbuild/protovalidate
+	buf generate --template conformance/buf.gen.yaml -o conformance/ buf.build/bufbuild/protovalidate-testing
+	./gradlew spotlessApply
```
pkwarren, line 65: We shouldn't run spotless on generated code - better to just exclude those generated files from spotless (this will also make the spotless checks run a lot faster).

buf.gen.yaml
```diff
@@ -0,0 +1,6 @@
+version: v1
+managed:
+  enabled: false
+plugins:
+  - plugin: buf.build/protocolbuffers/java:v23.4
```
pkwarren, line 5: To remove a dependency on the BSR as part of the build, you may wish to use https://github.com/google/protobuf-gradle-plugin and https://github.com/bufbuild/buf-gradle-plugin/ instead (which will allow you to configure code generation to work entirely locally).

conformance/buf.gen.yaml
```diff
@@ -0,0 +1,7 @@
+version: v1
+managed:
+  enabled: true
+  java_package_prefix: "build"
+plugins:
+  - plugin: buf.build/protocolbuffers/java:v23.4
```
pkwarren, line 6: Same here - I think we should switch to use local generation using buf/protobuf gradle plugins if we can.

conformance/src/main/java/build/buf/Main.java
```diff
@@ -88,13 +89,12 @@ static TestResult testCase(
...
-      if (result.getViolations().isEmpty()) {
+      List<Violation> violations = result.getViolations();
+      if (violations.isEmpty()) {
         return TestResult.newBuilder().setSuccess(true).build();
       } else {
-        return TestResult.newBuilder()
-            .setValidationError(
-                Violations.newBuilder().addAllViolations(result.getViolations()).build())
-            .build();
+        Violations error = Violations.newBuilder().addAllViolations(violations).build();
```
pkwarren, line 96: Small style thing (not from this change) but typically we shouldn't have an `else` after a `return` in the previous if block.

conformance/src/main/java/build/buf/validate/conformance/cases/AnEnum.java
```diff
@@ -0,0 +1,107 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
pkwarren, line 1: Please configure a .gitattributes file so that these are flagged as generated in follow ups: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

pkwarren (approved):
Looks great - nice work!

PR #22, opened 2023-08-21
Fixes #15: include .proto files in jar.

### LINKED ISSUE #15: Packaging validate.proto into jar
In [pgv](https://github.com/bufbuild/protoc-gen-validate), `validate.proto` is in the jar file. This allows for a better development experience with build tools (Gradle, Maven) and a way to look into the source definition.

image

hypnoce:


pkwarren:
Apologize - I didn't see that this existed already before creating #35. This will be resolved with that PR.

PR #24, opened 2023-08-23
Support for validation of isNaN & isInf

elliotmjackson:


src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -305,6 +307,27 @@ private static Overload isUriRef() {
...
+
+  /**
+   * Creates a custom unary function overload for the "isInf" operation.
+   *
+   * @return The {@link Overload} instance for the "isInf" operation.
+   */
+  private static Overload isInf() {
+    return Overload.unary(
+        "isInf",
+        value -> value.convertToNative(Double.TYPE).isInfinite() ? BoolT.True : BoolT.False);
```
Alfus, line 328: don't want to be compatible with all the other impls? :-)

Alfus (approved):

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
elliotmjackson: resolved in [cd1cb84](https://github.com/bufbuild/protovalidate-java/pull/24/commits/cd1cb84568bf8a3179dcfa45f2aca39752458458)

PR #36, opened 2023-09-12
Improve protovalidate overloads

pkwarren:
Update all of the protovalidate overload functions to add proper type checking and return Err.noSuchOverload if the types are incorrect. Greatly simplify the unique implementation as it didn't need to support non-list arguments. Fix a bug in the unique implementation for bytes (if the bytes contained invalid UTF-8). Add testcases for unique and additional tests for invalid arguments.

Update Format.format to use exceptions for control flow instead of having to distinguish Val.Err from other types (a lot of places were just returning null).

Remove all places where we used '==' to compare object type instances with comparing using enum types (which is safe for '==' comparison).

### DIFF
--- a/src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
+++ b/src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
@@ -26,25 +26,35 @@ import java.net.InetAddress;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
 import java.util.HashSet;
 import java.util.Set;
 import org.projectnessie.cel.common.types.BoolT;
-import org.projectnessie.cel.common.types.BytesT;
-import org.projectnessie.cel.common.types.DoubleT;
 import org.projectnessie.cel.common.types.Err;
 import org.projectnessie.cel.common.types.IntT;
 import org.projectnessie.cel.common.types.ListT;
 import org.projectnessie.cel.common.types.StringT;
 import org.projectnessie.cel.common.types.Types;
-import org.projectnessie.cel.common.types.UintT;
+import org.projectnessie.cel.common.types.ref.TypeEnum;
 import org.projectnessie.cel.common.types.ref.Val;
 import org.projectnessie.cel.common.types.traits.Lister;
 import org.projectnessie.cel.interpreter.functions.Overload;
-import org.projectnessie.cel.interpreter.functions.UnaryOp;
 
 /** Defines custom function overloads (the implementation). */
 final class CustomOverload {
+
+  private static final String OVERLOAD_FORMAT = "format";
+  private static final String OVERLOAD_UNIQUE = "unique";
+  private static final String OVERLOAD_STARTS_WITH = "startsWith";
+  private static final String OVERLOAD_ENDS_WITH = "endsWith";
+  private static final String OVERLOAD_CONTAINS = "contains";
+  private static final String OVERLOAD_IS_HOSTNAME = "isHostname";
+  private static final String OVERLOAD_IS_EMAIL = "isEmail";
+  private static final String OVERLOAD_IS_IP = "isIp";
+  private static final String OVERLOAD_IS_URI = "isUri";
+  private static final String OVERLOAD_IS_URI_REF = "isUriRef";
+  private static final String OVERLOAD_IS_NAN = "isNan";
+  private static final String OVERLOAD_IS_INF = "isInf";
+
   /**
    * Create custom function overload list.
    *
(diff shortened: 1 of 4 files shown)


src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -72,20 +82,20 @@ static Overload[] create() {
...
           ListT list = (ListT) rhs.convertToType(ListT.ListType);
-          String formatString = lhs.value().toString();
-          Val status = Format.format(formatString, list);
-          if (status.type() == Err.ErrType) {
-            return status;
+          String formatString = (String) lhs.value();
+          try {
+            return StringT.stringOf(Format.format(formatString, list));
+          } catch (Err.ErrException e) {
+            return e.getErr();
```
pkwarren, line 97: Updated this to use exceptions for control flow instead of dealing with interpreting the returned Val type (either Err or String).

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -94,29 +104,14 @@ private static Overload binaryFormat() {
...
-              }
-              Val firstValue = lister.get(IntT.intOf(0));
-              return unaryOpForPrimitiveVal(firstValue).invoke(lister);
-            case Bool:
-            case Bytes:
-            case Double:
-            case Int:
-            case String:
-            case Uint:
-              return unaryOpForPrimitiveVal(val).invoke(val);
```
pkwarren, line 116: A lot of these weren't valid for unique() - it should only operate on list types according to the Go implementation.

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -324,118 +359,62 @@ private static Overload isNan() {
...
-   */
-  private static Val uniqueScalar(Lister list) {
-    Set<Val> exist = new HashSet<>();
-    long size = list.size().intValue();
-    for (int i = 0; i < size; i++) {
+    exist.add(firstVal);
+    for (int i = 1; i < size; i++) {
       Val val = list.get(IntT.intOf(i));
-      if (exist.contains(val)) {
+      if (!exist.add(val)) {
```
pkwarren, line 415: The `add()` method returns false if the value is already contained in the set.

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -324,118 +359,62 @@ private static Overload isNan() {
...
-   *
-   * @param list The input list to check for uniqueness.
-   * @return {@link BoolT}.True if the list contains unique elements, {@link BoolT}.False otherwise.
-   */
-  private static Val uniqueBytes(Lister list) {
-    Set<Object> exist = new HashSet<>();
-    for (int i = 0; i < list.size().intValue(); i++) {
-      Object val = list.get(IntT.intOf(i)).value();
-      if (val instanceof byte[]) {
-        val = new String((byte[]) val, StandardCharsets.UTF_8);
```
pkwarren, line 411: This was the source of a bug if the bytes contained invalid UTF-8. If so, the string would contain replacement characters for invalid bytes and two  different bytes values would return equal.

src/main/java/build/buf/protovalidate/internal/celext/Format.java
```diff
@@ -91,37 +83,33 @@ static Val format(String fmtString, ListT list) {
...
           break;
         case 'e':
         case 'f':
         case 'b':
         case 'o':
         default:
-          return Err.newErr("format: unparsable format specifier %s", c);
-      }
-      if (status.type() == Err.ErrType) {
-        return status;
```
pkwarren, line 121: A lot of the complexity in here went away with switching to exceptions for control flow.

src/test/java/build/buf/protovalidate/internal/celext/CustomOverloadTest.java
```diff
@@ -63,6 +79,58 @@ public void testIsNan() {
...
+            .put("[1.0, 2.0, 3.0].unique()", true)
+            .put("[3.0,2.0,3.0].unique()", false)
+            .put("['abc', 'def'].unique()", true)
+            .put("['abc', 'abc'].unique()", false)
+            .put("[b'abc', b'123'].unique()", true)
+            .put("[b'123', b'123'].unique()", false)
+            // Previously, the unique() method returned false here as both bytes were converted
+            // to UTF-8. Since both contain invalid UTF-8, this would lead to them treated as equal
+            // because they'd have the same substitution character.
+            .put("[b'\\xFF', b'\\xFE'].unique()", true)
```
pkwarren, line 114: This failed before fixing the unique implementation (this test doesn't exist in the Go implementation but it doesn't suffer the same issues).

rodaine (approved):

PR #39, opened 2023-09-28
Add isIpPrefix

higebu:
Related to: https://github.com/bufbuild/protovalidate/pull/99


higebu:
Passed conformance tests with https://github.com/bufbuild/protovalidate/pull/99.

```console
$ buf export ../protovalidate/proto/protovalidate --output src/main/resources
$ buf generate --template buf.gen.yaml src/main/resources
$ buf generate --template conformance/buf.gen.yaml -o conformance/ ../protovalidate/proto/protovalidate-testing
$ ./gradlew build
$ ./gradlew conformance:jar
$ protovalidate-conformance --strict_message ./conformance/conformance.sh
PASS (failed: 0, skipped: 0, passed: 1100, total: 1100)
```

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -284,6 +288,53 @@ private static Overload isIp() {
...
+          String prefix = (String) lhs.value();
+          if (prefix.isEmpty()) {
+            return BoolT.False;
+          }
+          if (rhs.type().typeEnum() == TypeEnum.Int) {
+            return Types.boolOf(validateIPPrefix(prefix, rhs.intValue(), false));
+          } else if (rhs.type().typeEnum() == TypeEnum.Bool) {
+            return Types.boolOf(validateIPPrefix(prefix, 0L, rhs.booleanValue()));
+          }
+          return BoolT.False;
```
pkwarren, line 323: nit: Since we've already verified above the rhs is either Int/Bool, we can remove the `else if` and just `return Types.boolOf(validateIPPrefix(prefix, 0L, rhs.booleanValue()));`

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -284,6 +288,53 @@ private static Overload isIp() {
...
+        },
+        (values) -> {
+          if (values[0].type().typeEnum() != TypeEnum.String || values[1].type().typeEnum() != TypeEnum.Int || values[2].type().typeEnum() != TypeEnum.Bool) {
```
pkwarren, line 326: I don't think it is necessary but it might be good practice to check `values.length == 3` first before indexing into the values.

src/test/java/build/buf/protovalidate/internal/celext/CustomOverloadTest.java
```diff
@@ -131,6 +131,37 @@ public void testUniqueUnsupported() {
...
+            .put("'fd7a:115c:a1e0:ab12:4843:cd96:626b:430b/118'.isIpPrefix(6)", true)
+            .put("'fd7a:115c:a1e0:ab12:4843:cd96:626b:4000/118'.isIpPrefix(6,true)", true)
+            .put("'fd7a:115c:a1e0:ab12:4843:cd96:626b:430b/118'.isIpPrefix(6,true)", false)
+            .put("'1.2.3.0/24'.isIpPrefix(6)", false)
+            .build();
+    for (Map.Entry<String, Boolean> testCase : testCases.entrySet()) {
+      Program.EvalResult result = eval(testCase.getKey());
+      assertThat(result.getVal().booleanValue()).isEqualTo(testCase.getValue());
+    }
+  }
```
pkwarren, line 163: To match other overload tests, we should create an `testIsIpPrefixUnsupported` that calls `isIpPrefix` with invalid types/arguments.

pkwarren:
It also looks like we need to run `make lintfix` to get the CI jobs passing.

higebu:
@pkwarren Thank you for your review! I've addressed all your feedback.

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
```diff
@@ -284,6 +288,54 @@ private static Overload isIp() {
...
+          if (prefix.isEmpty()) {
+            return BoolT.False;
+          }
+          if (rhs.type().typeEnum() == TypeEnum.Int) {
+            return Types.boolOf(validateIPPrefix(prefix, rhs.intValue(), false));
+          }
+          return Types.boolOf(validateIPPrefix(prefix, 0L, rhs.booleanValue()));
+        },
+        (values) -> {
+          if (values.length != 3) {
```
pkwarren, line 327: I think we still want to check the types of all fields here as before, just wanted the length check as well first.

src/test/java/build/buf/protovalidate/internal/celext/CustomOverloadTest.java
```diff
@@ -131,6 +131,49 @@ public void testUniqueUnsupported() {
...
+
+  @Test
+  public void testIsIpPrefixUnsupported() {
+    List<String> testCases = ImmutableList.of("1.isIpPrefix()");
```
pkwarren, line 167: Can you add a few more cases here where we pass the wrong types of arguments to the function for one, two, and three (or more) argument versions?

src/main/java/build/buf/protovalidate/internal/celext/CustomOverload.java
higebu: I reverted checking types.

src/test/java/build/buf/protovalidate/internal/celext/CustomOverloadTest.java
higebu: I added test cases for the wrong types of arguments.

PR #45, opened 2023-10-10
Improve protovalidate with managed mode

pkwarren:
Update protovalidate to work with managed mode (without exceptions) and 'buf generate' as well as remote packages. Currently, those will generate references to the protovalidate extensions in locations other than the 'java_package' of 'build.buf.protovalidate' and when validation is performed, an IllegalArgumentException is thrown.

Update ConstraintResolver to use getField instead of getExtension to avoid the exception and transparently reparse extensions to the right classes to be used by protovalidate-java.

Add a test exercising this behavior which generates code using managed mode (without exceptions and '--include-imports') as well as one with managed mode (with an exception).


src/main/java/build/buf/protovalidate/internal/evaluator/ConstraintResolver.java
```diff
@@ -32,18 +34,24 @@ class ConstraintResolver {
...
    */
-  MessageConstraints resolveMessageConstraints(Descriptor desc) {
+  MessageConstraints resolveMessageConstraints(Descriptor desc)
+      throws InvalidProtocolBufferException {
     DescriptorProtos.MessageOptions options = desc.getOptions();
     if (!options.hasExtension(ValidateProto.message)) {
       return MessageConstraints.getDefaultInstance();
     }
-
-    MessageConstraints constraints = options.getExtension(ValidateProto.message);
```
pkwarren, line 41: This is the location where users were seeing:

```
mergeFrom(Message) can only merge messages of the same type.
java.lang.IllegalArgumentException: mergeFrom(Message) can only merge messages of the same type.
```

This occurred when the generated .java files from protoc referenced protovalidate extensions in another location (i.e. `build.buf.gen.buf.validate.*` for remote packages) while we tried to read them as `build.buf.validate.*` classes.

We should still resolve this in remote packages to avoid having duplicate classes generated to different Java packages on the classpath, but this at least avoids the exception and also has the side effect of working for users that use `buf generate` with managed mode.

src/main/java/build/buf/protovalidate/internal/evaluator/ConstraintResolver.java
```diff
@@ -32,18 +34,24 @@ class ConstraintResolver {
...
+    // Don't use getExtension here to avoid exception if descriptor types don't match.
+    // This can occur if the extension is generated to a different Java package.
+    Object value = options.getField(ValidateProto.message.getDescriptor());
+    if (value instanceof MessageConstraints) {
+      return ((MessageConstraints) value);
+    }
+    if (value instanceof MessageLite) {
+      // Possible that this represents the same constraint type, just generated to a different
+      // java_package.
+      return MessageConstraints.parseFrom(((MessageLite) value).toByteString());
```
pkwarren, line 52: Here we take the extension which exists as another Java class in another package and reparse it to the one used by this library.

It is possible that if versions are different between the two (i.e. older protovalidate-java with newer bufbuild/protovalidate generated code) that there will be unknown fields, however that's better than failing with an exception IMO and is easily fixed by updating protovalidate-java to a newer version.

src/test/java/build/buf/protovalidate/ValidatorDifferentJavaPackagesTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+import build.buf.protovalidate.exceptions.ValidationException;
+import build.buf.validate.Violation;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
+import java.util.Collections;
+import java.util.List;
+import org.junit.Test;
+
+/**
```
pkwarren, line 27: This provides full details on what we're fixing and tests both types of generated code and verifies it gets the same results.

src/main/java/build/buf/protovalidate/internal/evaluator/ConstraintResolver.java
```diff
@@ -32,18 +34,24 @@ class ConstraintResolver {
...
     }
-    return constraints;
+    return MessageConstraints.getDefaultInstance();
```
Alfus, line 54: Seems like this should never happen, and thus maybe be an exception, instead of silently dropping constraints but using the default instance.

Alfus (approved):

src/main/java/build/buf/protovalidate/internal/evaluator/ConstraintResolver.java
pkwarren: Good call - this branch was unreachable in the tests and better to throw an exception to catch sooner than transparently fail to perform validation.

rodaine (approved):

PR #46, opened 2023-10-11
Migrate build steps to Gradle build

pkwarren:
Currently, certain Gradle tasks (like 'build', 'check') don't work without running 'make generate' first. Instead of splitting logic in between Gradle and the Makefile, migrate logic for generating code, running license-header, and other details to the Gradle build so they can happen in the correct order automatically. This has the side effect that new users can use Gradle commands they are familiar with ('./gradlew build') and not have to learn the makefile conventions separately.


.github/workflows/conformance.yaml
```diff
@@ -24,7 +24,5 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
           cache: 'gradle'
-      - name: Generate
-        run: make checkgenerate
```
pkwarren, line 28: I had to add this step in the previous PR because it would fail unless the generation step is done before. This would also break the release flow. By moving the logic and dependencies to the Gradle build, we can ensure that prior to compiling Java code we run necessary code generation steps first.

Makefile
```diff
@@ -33,32 +25,19 @@ checkgenerate: generate  ## Checks if `make generate` produces a diff.
 
 .PHONY: clean
 clean:  ## Delete intermediate build artifacts
-	@# -X only removes untracked files, -d recurses into directories, -f actually removes files/dirs
-	git clean -Xdf
```
pkwarren, line 37: This was removing the `.idea/` directory used to work on this project too. By switching to Gradle (and ensuring all created files are created in `build/`), we can just use the normal clean step.

build.gradle.kts
```diff
@@ -18,9 +18,108 @@ java {
...
+    commandLine(bufCLIPath, "generate", "--template", "buf.gen.yaml", "src/main/resources")
+}
+
+tasks.register<Exec>("generateConformance") {
+    dependsOn("installBuf")
+    description = "Generates sources for the bufbuild/protovalidate-testing module to conformance/src/main/java."
+    commandLine(bufCLIPath, "generate", "--template", "buf.gen.yaml", "src/main/resources")
+}
+
+tasks.register("generate") {
```
pkwarren, line 106: This replaces the steps from the `make generate` target in the Makefile.

build.gradle.kts
```diff
@@ -18,9 +18,108 @@ java {
...
+        "generateSources",
+        "generateConformance",
+        "licenseHeader",
+    )
+}
+
 tasks.withType<JavaCompile> {
-    if (JavaVersion.current().isJava9Compatible) doFirst {
-        options.compilerArgs = mutableListOf("--release", "8")
+    dependsOn("generateTestSources")
```
pkwarren, line 118: This ensures that we've generated the source code necessary to compile the tests prior to compiling code.

build.gradle.kts
```diff
@@ -60,7 +159,11 @@ sourceSets {
 apply(plugin = "com.diffplug.spotless")
 configure<SpotlessExtension> {
     java {
-        targetExclude("src/main/java/build/buf/validate/**/*.java")
+        targetExclude("src/main/java/build/buf/validate/**/*.java", "build/generated/test-sources/bufgen/**/*.java")
+    }
+    kotlinGradle {
```
pkwarren, line 164: Now running spotless to auto format the Gradle build files.

conformance/build.gradle.kts
```diff
@@ -4,13 +4,38 @@ import net.ltgt.gradle.errorprone.errorprone
...
+    environment("GOBIN", conformanceCLIFile.parentFile.absolutePath)
+    outputs.file(conformanceCLIFile)
+    commandLine(
+        "go",
+        "install",
+        "github.com/bufbuild/protovalidate/tools/protovalidate-conformance@${project.findProperty("protovalidate.version")}",
+    )
+}
+
+tasks.register<Exec>("conformance") {
```
pkwarren, line 28: This logic was specific to this subdirectory (and project) but was included in the top-level makefile before. It helps to move it closer to where it is used.

conformance/src/main/java/build/buf/protovalidate/conformance/FileDescriptorUtil.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package build.buf;
+package build.buf.protovalidate.conformance;
```
pkwarren, line 15: Moved this to a better location specific to this project.

build.gradle.kts
```diff
@@ -94,15 +197,16 @@ mavenPublishing {
     val releaseVersion = project.findProperty("releaseVersion") as String? ?: System.getenv("VERSION")
```
stefanvanburen, line 197: not really related to this PR, but I'm seeing:

```
> Configure project :
w: file:///.../build.gradle.kts:214:13: Name shadowed: releaseVersion
```

I'm assuming that's deliberate so that we don't pick up whatever `VERSION` is in the environment down below in `pom {`, but just sanity checking.

build.gradle.kts
```diff
@@ -18,9 +18,108 @@ java {
...
+tasks.register<Exec>("generateSources") {
+    dependsOn("installBuf")
+    description = "Generates sources for the bufbuild/protovalidate module sources to src/main/java."
+    commandLine(bufCLIPath, "generate", "--template", "buf.gen.yaml", "src/main/resources")
+}
+
+tasks.register<Exec>("generateConformance") {
+    dependsOn("installBuf")
+    description = "Generates sources for the bufbuild/protovalidate-testing module to conformance/src/main/java."
+    commandLine(bufCLIPath, "generate", "--template", "buf.gen.yaml", "src/main/resources")
```
stefanvanburen, line 103: I might be missing it, but this looks like the same `buf generate` call as on line 97 - I would think these would be different?

in the makefile they are:

```
	$(BIN)/buf generate --template buf.gen.yaml src/main/resources
	$(BIN)/buf generate --template conformance/buf.gen.yaml -o conformance/ buf.build/bufbuild/protovalidate-testing:$(PROTOVALIDATE_VERSION)
```

stefanvanburen (commented):
looks reasonable to me, one nit and one question

build.gradle.kts
pkwarren: Good eye - fixed. Verified by removing the generated code and re-running `./gradlew generate` that it re-added it.

build.gradle.kts
pkwarren: I think that warning has been there for some time - in a follow up I'd like to make some changes here similar to changes in connect-kotlin to better choose release versions based on the most recent git tag, etc. I can look more at this error as part of that work.

pkwarren: Fixed the shadowing error as there was another typo in the name. Verified that local release works after the changes.

stefanvanburen (approved):

PR #48, opened 2023-10-16
Improve support for dynamic messages

pkwarren:
Update protovalidate-java to determine if the field, message, or oneof options contains an unknown field for the protovalidate extension. If so, reparse the options type to correctly interpret the options and enable validation.

This will enable protovalidate to run when the inputs are a FileDescriptorSet (with preserved options).

### DIFF
--- a/src/main/java/build/buf/protovalidate/exceptions/CompilationException.java
+++ b/src/main/java/build/buf/protovalidate/exceptions/CompilationException.java
@@ -14,12 +14,13 @@
 
 package build.buf.protovalidate.exceptions;
 
-/**
- * {@link CompilationException} extends {@link ValidationException} is returned when a constraint
- * fails to compile. This is a fatal error.
- */
+/** CompilationException is returned when a constraint fails to compile. This is a fatal error. */
 public class CompilationException extends ValidationException {
   public CompilationException(String message) {
     super(message);
+  }
+
+  public CompilationException(String message, Throwable cause) {
+    super(message, cause);
   }
 }
--- a/src/main/java/build/buf/protovalidate/internal/evaluator/ConstraintResolver.java
+++ b/src/main/java/build/buf/protovalidate/internal/evaluator/ConstraintResolver.java
@@ -23,6 +23,7 @@ import com.google.protobuf.DescriptorProtos;
 import com.google.protobuf.Descriptors.Descriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.Descriptors.OneofDescriptor;
+import com.google.protobuf.ExtensionRegistry;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.MessageLite;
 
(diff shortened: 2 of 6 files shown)


src/main/java/build/buf/protovalidate/exceptions/CompilationException.java
```diff
@@ -14,12 +14,13 @@
...
- * {@link CompilationException} extends {@link ValidationException} is returned when a constraint
- * fails to compile. This is a fatal error.
- */
+/** CompilationException is returned when a constraint fails to compile. This is a fatal error. */
 public class CompilationException extends ValidationException {
   public CompilationException(String message) {
     super(message);
   }
+
+  public CompilationException(String message, Throwable cause) {
```
pkwarren, line 23: Added overridden constructor so if we have an Exception cause, we can preserve it for debugging.

src/main/java/build/buf/protovalidate/internal/evaluator/ConstraintResolver.java
```diff
@@ -35,9 +36,13 @@ class ConstraintResolver {
    * @param desc the message descriptor.
    * @return the resolved {@link MessageConstraints}.
    */
-  MessageConstraints resolveMessageConstraints(Descriptor desc)
+  MessageConstraints resolveMessageConstraints(Descriptor desc, ExtensionRegistry registry)
       throws InvalidProtocolBufferException, CompilationException {
     DescriptorProtos.MessageOptions options = desc.getOptions();
+    // If the protovalidate message extension is unknown, reparse using extension registry.
+    if (options.getUnknownFields().hasField(ValidateProto.message.getNumber())) {
+      options = DescriptorProtos.MessageOptions.parseFrom(options.toByteString(), registry);
```
pkwarren, line 44: This is the key fix for handling dynamic messages created from a file descriptor set. If we determine that the protovalidate extension is in the unknown fields, we'll reparse it using our extension registry.

src/test/java/build/buf/protovalidate/ValidatorDynamicMessageTest.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * This test mimics the behavior when performing validation with protovalidate on a file descriptor
+ * set (as created by <code>protoc --retain_options --descriptor_set_out=...</code>). These
+ * descriptor types have the protovalidate extensions as unknown fields and need to be parsed with
+ * an extension registry for the constraints to be recognized and validated.
+ */
+public class ValidatorDynamicMessageTest {
+
+  @Test
+  public void testFieldConstraintDynamicMessage() throws Exception {
```
pkwarren, line 44: All of these tests previously failed for field, message, and oneof constraints.

src/test/java/build/buf/protovalidate/ValidatorDynamicMessageTest.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+    Descriptors.FileDescriptor[] dependencies =
+        new Descriptors.FileDescriptor[fdProto.getDependencyCount()];
+    for (int i = 0; i < fdProto.getDependencyCount(); i++) {
+      dependencies[i] = getFileDescriptor(fdProto.getDependency(i), fds);
+    }
+    return Descriptors.FileDescriptor.buildFrom(fdProto, dependencies);
+  }
+
+  private static DynamicMessage.Builder createMessageWithUnknownOptions(Message message)
```
pkwarren, line 122: This is just a round trip operation to go from a message created from Java generated code -> FileDescriptorSet -> DynamicMessage.Builder. The key thing is on line 126-127 where we reparse the FileDescriptorSet from binary format (losing any resolved extensions, etc.).

Alfus (approved):

PR #53, opened 2023-10-20
Add timestamp.valid evaluator

simonbos:
See https://github.com/bufbuild/protovalidate/issues/101 for details.

This depends on https://github.com/bufbuild/protovalidate/pull/114 being merged. However, locally I was able to build & run conformance tests using some edits to the `build.gradle.kts` and `Makefile`. (After the merge on the main repo, I would best re-generate the generated stubs.)

### DIFF
--- a/src/main/java/build/buf/protovalidate/internal/evaluator/TimestampEvaluator.java
+++ b/src/main/java/build/buf/protovalidate/internal/evaluator/TimestampEvaluator.java
@@ -1,0 +1,88 @@
+// Copyright 2023 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buf.protovalidate.internal.evaluator;
+
+import build.buf.protovalidate.ValidationResult;
+import build.buf.protovalidate.exceptions.ExecutionException;
+import build.buf.validate.Violation;
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.Message;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A specialized evaluator for applying some {@link build.buf.validate.TimestampRules} (only the
+ * `valid` rule currently) to an {@link com.google.protobuf.Timestamp} message. This is handled
+ * outside CEL which handles {@link com.google.protobuf.Timestamp} as an abstract type, thus not
+ * allowing access to the message fields.
+ */
+class TimestampEvaluator implements Evaluator {
+  private final long maxTimestamp = +253402300799L;
+  private final long minTimestamp = -62135596800L;
+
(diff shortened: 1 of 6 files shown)


src/main/java/build/buf/protovalidate/internal/evaluator/TimestampEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+  private final Descriptors.FieldDescriptor secondsDescriptor;
+  private final Descriptors.FieldDescriptor nanosDescriptor;
+  private final boolean valid;
+
+  /** Constructs a new evaluator for {@link build.buf.validate.TimestampRules} messages. */
+  TimestampEvaluator(
+      Descriptors.FieldDescriptor secondsDescriptor,
+      Descriptors.FieldDescriptor nanosDescriptor,
+      boolean valid) {
+    this.secondsDescriptor = secondsDescriptor;
```
pkwarren, line 44:
```suggestion
    this.secondsDescriptor = Objects.requireNonNull(secondsDescriptor, "secondsDescriptor");
```

src/main/java/build/buf/protovalidate/internal/evaluator/TimestampEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+    this.nanosDescriptor = nanosDescriptor;
```
pkwarren, line 45:
```suggestion
    this.nanosDescriptor = Objects.requireNonNull(nanosDescriptor, "nanosDescriptor");
```

src/main/java/build/buf/protovalidate/internal/evaluator/TimestampEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+      } else if (nanos < 0 || nanos >= 1e9) {
+        errorMessage = "timestamp has out-of-range nanos";
+      }
+
+      if (errorMessage.length() != 0) {
+        Violation violation =
+            Violation.newBuilder()
+                .setConstraintId("timestamp.valid")
+                .setMessage(errorMessage)
+                .build();
```
pkwarren, line 74: We should set the field path on the violation as well.

src/main/java/build/buf/protovalidate/internal/evaluator/TimestampEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+        violationList.add(violation);
+        if (failFast) {
```
pkwarren, line 76: This isn't looping over anything here, so I think we can just remove this if statement.

src/main/java/build/buf/protovalidate/internal/evaluator/TimestampEvaluator.java
```diff
@@ -0,0 +1,88 @@
...
+      long seconds = (long) timestampValue.getField(secondsDescriptor);
+      int nanos = (int) timestampValue.getField(nanosDescriptor);
+
+      String errorMessage = "";
+      if (seconds < minTimestamp) {
+        errorMessage = "timestamp before 0001-01-01";
+      } else if (seconds > maxTimestamp) {
+        errorMessage = "timestamp after 9999-12-31";
+      } else if (nanos < 0 || nanos >= 1e9) {
+        errorMessage = "timestamp has out-of-range nanos";
```
pkwarren, line 66: I think it would be useful if the error message included details of all parts of the timestamp which are invalid. So if both the seconds and nanos values are invalid, we should return that in the error message.

Do we also want to include in the error message the expected range of the timestamp nanos?

pkwarren (commented):
Changes look good - just had a few minor comments. I know this is waiting for upstream PRs to be unblocked.

PR #60, opened 2023-11-08
Update required/ignored_empty behavior to match spec

rodaine:
Bringing the `required` and `ignore_empty` constraints into conformance.

See https://github.com/bufbuild/protovalidate/issues/115


src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBuilder.java
```diff
@@ -193,6 +191,36 @@ private void buildValue(
...
+            .addField(desc.toProto().toBuilder().clearLabel());
+        Descriptors.FileDescriptor[] deps = {};
+        try {
+          desc =
+              Descriptors.FileDescriptor.buildFrom(bldr.build(), deps)
+                  .findMessageTypeByName(desc.getContainingType().getName())
+                  .findFieldByNumber(desc.getNumber());
+        } catch (Descriptors.DescriptorValidationException e) {
+          throw new CompilationException(e.toString());
+        }
```
rodaine, line 218: @pkwarren this is the ☠️ I was talking about. I could not find any other way to get the zero value of a repeated field's elements.

src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBuilder.java
```diff
@@ -27,12 +27,9 @@
 import build.buf.validate.MessageConstraints;
 import build.buf.validate.OneofConstraints;
 import build.buf.validate.ValidateProto;
-import com.google.protobuf.Descriptors;
+import com.google.protobuf.*;
```
pkwarren, line 30: I think we should update our code style to prevent wildcard imports for consistency across projects.

src/main/java/build/buf/protovalidate/internal/evaluator/MessageValue.java
```diff
@@ -56,4 +56,9 @@ public List<Value> repeatedValue() {
   public Map<Value, Value> mapValue() {
     return Collections.emptyMap();
   }
+
+  @Override
+  public Object raw() {
```
pkwarren, line 61: Instead of this method, we should be able to call `Value.value(Object.class)` instead.

src/main/java/build/buf/protovalidate/internal/evaluator/ValueEvaluator.java
```diff
@@ -16,23 +16,18 @@
...
 import build.buf.protovalidate.ValidationResult;
 import build.buf.protovalidate.exceptions.ExecutionException;
-import build.buf.validate.FieldConstraints;
 import build.buf.validate.Violation;
-import com.google.protobuf.Descriptors;
-import com.google.protobuf.DynamicMessage;
 import java.util.ArrayList;
 import java.util.List;
-import org.projectnessie.cel.common.ULong;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
pkwarren, line 22: For consistency, I think we want to use `javax.annotation.Nullable`.

src/main/java/build/buf/protovalidate/internal/evaluator/ValueEvaluator.java
```diff
@@ -108,13 +76,12 @@ public void append(Evaluator eval) {
...
-      return val.value(Object.class) == null;
-    }
-    return zero.equals(val.value(zero.getClass()));
+  public void setIgnoreEmpty(Object zero) {
+    this.ignoreEmpty = true;
+    this.zero = zero;
+  }
+
+  private boolean shouldIgnore(Object value) {
+    return this.ignoreEmpty && value == this.zero;
```
pkwarren, line 85:
```suggestion
    return this.ignoreEmpty && Objects.equals(value, this.zero);
```

pkwarren (approved):

PR #62, opened 2023-11-10
Update EvaluatorBuilder cache to be thread safe

### LINKED ISSUE #50: [BUG] Validator/EvaluatorBuilder aren't thread safe
## Description
Currently, the Validator instance isn't thread-safe. The collections it uses (via the EvaluatorBuilder) are all meant to be accessed/modified from a single thread (ArrayList, HashMap) and as written sharing a Validator by multiple threads isn't safe.

## Steps to Reproduce
n/a - Found via code inspection. This might manifest itself as collection issues, infinite loops (in some older versions of Java these could be triggered), or incorrect cached metadata for a constraint.

## Expected Behavior
A Validator instance should be safe for use across multiple threads (so it can benefit from caching).

## Actual Behavior
A Validator instance is only safe when accessed from a single thread.

## Screenshots/Logs
n/a

## Environment

- **Operating System**: macOS
- **Version**: macOS 14.0
- **Compiler/Toolchain**: Java 17
- **Protobuf Compiler & Version**: 3.24.4
- **Protoc-gen-validate Version**: n/a
- **Protovalidate Version**: v0.1.6

## Possible Solution
At a minimum the `EvaluatorBuilder.load(...)` method needs to have the `synchronized` keyword. Alternatively, thread-safe collections should be used (and we may need to rethink the algorithm for loading the cache as we first load a partial message's constraints in the map before it is fully initialized).

## Additional Context
n/a

pkwarren:
Update EvaluatorBuilder to add locking around cache misses to avoid concurrent modification to a HashMap. Migrate all of the code to build a new cache to an inner class to better separate code concerns.

Fixes #50.

### DIFF
--- a/src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBuilder.java
+++ b/src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBuilder.java
@@ -27,6 +27,7 @@ import build.buf.validate.FieldConstraints;
 import build.buf.validate.MessageConstraints;
 import build.buf.validate.OneofConstraints;
 import build.buf.validate.ValidateProto;
+import com.google.common.collect.ImmutableMap;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Descriptors.Descriptor;
@@ -39,7 +40,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.Objects;
 import org.projectnessie.cel.Env;
 import org.projectnessie.cel.EnvOption;
 import org.projectnessie.cel.checker.Decls;
@@ -54,8 +55,7 @@ public class EvaluatorBuilder {
     EXTENSION_REGISTRY.add(ValidateProto.oneof);
   }
 
-  private final Map<Descriptor, Evaluator> evaluatorMap = new HashMap<>();
-  private final ConstraintResolver resolver = new ConstraintResolver();
+  private volatile ImmutableMap<Descriptor, Evaluator> evaluatorCache = ImmutableMap.of();
 
   private final Env env;
   private final boolean disableLazy;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBuilder.java
```diff
@@ -78,358 +78,404 @@ public EvaluatorBuilder(Env env, boolean disableLazy) {
...
-            EnvOption.types(message),
-            EnvOption.declarations(
-                Decls.newVar(Variable.THIS_NAME, Decls.newObjectType(desc.getFullName()))));
-    List<CompiledProgram> compiledPrograms = compileConstraints(celList, finalEnv);
-    if (compiledPrograms.isEmpty()) {
-      throw new CompilationException("compile returned null");
-    }
-    msgEval.append(new CelPrograms(compiledPrograms));
-  }
+  private static class DescriptorCacheBuilder {
```
pkwarren, line 117: I'd recommend viewing the rest of this ignoring whitespace - it is really just moving functions inside this inner static class.

src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBuilder.java
```diff
@@ -78,358 +78,404 @@ public EvaluatorBuilder(Env env, boolean disableLazy) {
...
-  private void buildMessage(Descriptor desc, MessageEvaluator msgEval) throws CompilationException {
-    try {
-      DynamicMessage defaultInstance =
-          DynamicMessage.newBuilder(desc).mergeFrom(new byte[0], EXTENSION_REGISTRY).buildPartial();
-      Descriptor descriptor = defaultInstance.getDescriptorForType();
-      MessageConstraints msgConstraints =
-          resolver.resolveMessageConstraints(descriptor, EXTENSION_REGISTRY);
-      if (msgConstraints.getDisabled()) {
-        return;
+    synchronized (this) {
```
pkwarren, line 97: This is the key change - if we have a cache miss we'll check again (in case another thread beat us in the meantime) and if that fails, we'll rebuild a brand new cache map from the current one and the descriptor.

src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBuilder.java
```diff
@@ -54,8 +55,7 @@ public class EvaluatorBuilder {
     EXTENSION_REGISTRY.add(ValidateProto.oneof);
   }
 
-  private final Map<Descriptor, Evaluator> evaluatorMap = new HashMap<>();
-  private final ConstraintResolver resolver = new ConstraintResolver();
+  private volatile ImmutableMap<Descriptor, Evaluator> evaluatorCache = ImmutableMap.of();
```
pkwarren, line 58: I don't have to use immutablemap here, but I thought it was a good indication to make it clear that this cache won't be modified - if we have a cache miss we'll create a new map.

Alfus (approved):

PR #75, opened 2023-12-18
Fix integer overflow bug in validation error message

lyang:
## Symptom
Some standard validation error messages have `Integer Overflow` bug on large decimal values.

## Example
```protobuf
message LargeValue {
  uint64 count = 1 [(buf.validate.field).uint64 = {gte: 0, lte: 999999999999}];
}
```

```java
@Test
public void validate_count() {
  LargeValue value = LargeValue.newBuilder().setCount(9999999999999).build();
  ValidationResult result = new Validator().validate(value);
  assertThat(result.getViolations().get(0).getMessage()).isEqualTo("value must be greater than or equal to 0 and less than or equal to 999999999999");
  // Above fails with: "value must be greater than or equal to 0 and less than or equal to -727379969"
}
```

## Cause
This is due to `%s` interpreted as `String` in [`Format.java`](https://github.com/bufbuild/protovalidate-java/blob/b35a45f8eccb0aef7d5a0fdec20d0a4f159529f7/src/main/java/build/buf/protovalidate/internal/celext/Format.java#L101-L103):
```java
// ...
        case 's':
          formatString(builder, arg);
          break;
// ...
```
which eventually calls:
```java
// ...
    } else if (type == TypeEnum.Int || type == TypeEnum.Uint) {
      formatInteger(builder, Long.valueOf(val.intValue()).intValue());
// ...
```

## Note
Originally fixed in a different way in https://github.com/bufbuild/protovalidate/pull/144

### DIFF
--- a/src/main/java/build/buf/protovalidate/internal/celext/Format.java
+++ b/src/main/java/build/buf/protovalidate/internal/celext/Format.java
@@ -248,20 +248,6 @@ final class Format {
   }
 
   /**
-   * Formats an integer value.
-   *
-   * @param builder the StringBuilder to append the formatted integer value to.
-   * @param value the value to format.
-   */
-  private static void formatInteger(StringBuilder builder, int value) {
-    if (value < 0) {
-      builder.append("-");
-      value = -value;
-    }
-    builder.append(value);
-  }
-
-  /**
    * Formats a hexadecimal value.
    *
    * @param builder the StringBuilder to append the formatted hexadecimal value to.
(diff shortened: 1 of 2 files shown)


src/main/java/build/buf/protovalidate/internal/celext/Format.java
```diff
@@ -157,7 +157,7 @@ private static void formatStringSafe(StringBuilder builder, Val val, boolean lis
     if (type == TypeEnum.Bool) {
       builder.append(val.booleanValue());
     } else if (type == TypeEnum.Int || type == TypeEnum.Uint) {
-      formatInteger(builder, Long.valueOf(val.intValue()).intValue());
```
nicksnyder, line 160: If formatInteger is no longer used anywhere, can you delete the implementation?

lyang-highnote: Sure, will do

pkwarren (approved):

lyang:
Hi,
Just want to check in and see if there's a release scheduled soon?

Thanks!

nicksnyder:
Just released [0.1.9](https://github.com/bufbuild/protovalidate-java/releases/tag/v0.1.9)

PR #107, opened 2024-03-05
Update copyright year range

pkwarren:


stefanvanburen (approved):

drice-buf:
Is this actually a necessary thing to do from a legal standpoint?

pkwarren:
> Is this actually a necessary thing to do from a legal standpoint?

By convention we do this in each OSS repo.

drice-buf (approved):

PR #108, opened 2024-03-05
CI improvements

pkwarren:
Update gradle to the latest version, build in CI with the latest LTS Java version, and remove unused Maven repository for generated SDKs (code generation has moved to happen in this project directly).


.github/workflows/ci.yaml
```diff
@@ -17,17 +17,22 @@ jobs:
...
+      - name: Cache Go Modules
+        uses: actions/cache@v4
         with:
-          fetch-depth: 1
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-gomod-ci-${{ hashFiles('gradle.properties', 'gradle/libs.versions.toml') }}
+          restore-keys:
+            ${{ runner.os }}-gomod-ci-
```
pkwarren, line 28: For the ci/release workflows, we now cache go modules and build for building the license header binary. We no longer need to build the buf binary as we pull it from the Maven artifact.

.github/workflows/ci.yaml
```diff
@@ -17,17 +17,22 @@ jobs:
...
       - uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: '21'
```
pkwarren, line 32: Bumping to latest LTS - we still build with source/target/release set to Java 8 compatibility.

.github/workflows/conformance.yaml
```diff
@@ -17,12 +17,21 @@ jobs:
...
+      - name: Cache Go Modules
+        uses: actions/cache@v4
         with:
-          fetch-depth: 1
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-gomod-conformance-${{ hashFiles('gradle.properties', 'gradle/libs.versions.toml') }}
+          restore-keys:
+            ${{ runner.os }}-gomod-conformance-
```
pkwarren, line 28: Conformance workflow uses a different cache as it also downloads and builds the conformance binary from bufbuild/protovalidate.

build.gradle.kts
```diff
@@ -31,23 +32,20 @@ if (matchResult != null) {
...
 val bufLicenseHeaderCLIFile = project.layout.buildDirectory.file("gobin/license-header").get().asFile
 val bufLicenseHeaderCLIPath: String = bufLicenseHeaderCLIFile.absolutePath
 
-tasks.register<Exec>("installBuf") {
+tasks.register("configureBuf") {
     description = "Installs the Buf CLI."
-    environment("GOBIN", bufCLIFile.parentFile.absolutePath)
-    outputs.file(bufCLIFile)
-    commandLine("go", "install", "github.com/bufbuild/buf/cmd/buf@latest")
+    File(buf.asPath).setExecutable(true)
```
pkwarren, line 41: Instead of needing to download all the buf dependencies and build the CLI, we just download the Maven artifact and make it executable. This also has the benefit of including the artifact in the existing gradle CI cache.

build.gradle.kts
```diff
@@ -31,23 +32,20 @@ if (matchResult != null) {
...
 }
 
 tasks.register<Exec>("installLicenseHeader") {
     description = "Installs the Buf license-header CLI."
     environment("GOBIN", bufLicenseHeaderCLIFile.parentFile.absolutePath)
     outputs.file(bufLicenseHeaderCLIFile)
-    commandLine("go", "install", "github.com/bufbuild/buf/private/pkg/licenseheader/cmd/license-header@latest")
+    commandLine("go", "install", "github.com/bufbuild/buf/private/pkg/licenseheader/cmd/license-header@v${libs.versions.buf.get()}")
```
pkwarren, line 48: This should keep our builds stable as we no longer depend on latest.

build.gradle.kts
```diff
@@ -193,14 +191,10 @@ allprojects {
     version = releaseVersion
     repositories {
         mavenCentral()
-        maven {
-            name = "buf"
-            url = uri("https://buf.build/gen/maven")
-        }
```
pkwarren, line 199: Originally this project used a generated SDK, but it now builds the conformance protos as part of this project so we can remove this.

build.gradle.kts
```diff
@@ -276,9 +270,12 @@ dependencies {
     implementation(libs.ipaddress)
     implementation(libs.jakarta.mail.api)
 
+    buf("build.buf:buf:${libs.versions.buf.get()}:${osdetector.classifier}@exe")
+
     testImplementation(libs.assertj)
     testImplementation(platform(libs.junit.bom))
     testImplementation("org.junit.jupiter:junit-jupiter")
+    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
```
pkwarren, line 278: Fixes a gradle warning.

gradle/libs.versions.toml
```diff
@@ -10,6 +11,7 @@ protobuf = "3.25.3"
 
 [libraries]
 assertj = { module = "org.assertj:assertj-core", version.ref = "assertj" }
+buf = { module = "buf.build:buf", version.ref = "buf" }
```
pkwarren, line 14: 🤞 that this will keep the Buf CLI updated with dependabot.

build.gradle.kts
jhump: 👍. I really _wanted_ to use a generated SDK for the newer conformance protos, but then we don't have the option to build for both lite and standard Java runtimes. Do we? If there's a way to pick which runtime to use in the URL, I'm all ears and will happily move this back to using generated SDKs (since that's what we use in conformance tests for other languages that generated SDKs support).

jhump (approved):

build.gradle.kts
pkwarren: Yep - you can use a `lite` runtime artifact for plugins that support it: https://buf.build/docs/bsr/generated-sdks/maven#remote

PR #111, opened 2024-03-07
Fix buf CLI group ID

pkwarren:


gradle/libs.versions.toml
```diff
@@ -11,7 +11,7 @@ protobuf = "3.25.3"
 
 [libraries]
 assertj = { module = "org.assertj:assertj-core", version.ref = "assertj" }
-buf = { module = "buf.build:buf", version.ref = "buf" }
+buf = { module = "build.buf:buf", version.ref = "buf" }
```
pkwarren, line 14: Had added this here just so dependabot could keep buf CLI updated for us, however it isn't used as a dependency (since this is just the .pom artifact). This is where we use the version that we want to keep updated: https://github.com/bufbuild/protovalidate-java/blob/bbee79919fd1ff85cc9b69a605aeea05fe3b78a7/build.gradle.kts#L273

jhump (approved):

Issue #80, opened 2023-12-26
[Feature Request] Extending protovalidate-java to other JVM Protobuf implementations


andrewparmet:
**Feature description:** 

I'd like to generify the `Value` interface to accommodate validation of Protobuf messages generated by other implementations. Specifically I'd like to implement support for [protokt](https://github.com/open-toast/protokt).

**Problem it solves or use case:** 

It would be nice to write another implementation of protovalidate for a JVM language without having to reinvent the wheel or fork this project.

**Proposed implementation or solution:** 

It turns out you don't have to do much to generify this project.

First you change `Value` to reference a generic "MessageLike" interface. Then from everything that fails to compile, I derived the exact reflection capabilities protovalidate-java requires of a Protobuf runtime:

```jave
// this name is subject to change
interface MessageLike {
    // Generify ObjectValue creation so an implementation can return its own types.
    // This could be more enterprisey but it gets the job done.
    Value newObjectValue(FieldDescriptor fieldDescriptor, Object fieldValue: Object);

    // Get the count of a repeated field on a message.
    int getRepeatedFieldCount(FieldDescriptor field);

    // Whether a message has a field. This can be a standard field or a member of a oneof.
    boolean hasField(FieldDescriptor field);

    // Whether a message has a oneof populated.
    boolean hasField(OneofDescriptor oneof);

    // Get the value of a field. This can be a standard field or a member of a oneof.
    Object getField(FieldDescriptor field);
}
```

**Contribution:**

I have already developed an implementation of this proposal. My forked version of protovalidate-java passes all conformance tests for protobuf-java and protokt. My branch is a bit sloppy since it's a POC, but the conformance tests can be swapped from the Java implementation to the protokt implementation by changing the main class of the `conformance:conformance` task from `Main` to `Main2`.

https://github.com/andrewparmet/protovalidate-protokt/pull/9

If my changes to the various evaluator classes can be merged here, then I can write a spinoff library that reuses the majority of the implementation code in this project to validate Protokt messages.

As a bonus it would be great to be able to generate the conformance test cases using the protobuf-gradle-plugin from a JAR or something so I don't have to copy them manually - or I can write a task to download them from a Buf repo if they're available that way stably.

Thanks!

**Additional context:** 
I'm the author of protokt and the buf-gradle-plugin.

PR #132, opened 2024-05-12
Introduce MessageReflector to enable alternative protobuf runtimes

andrewparmet:
see https://github.com/bufbuild/protovalidate-java/issues/80

see https://github.com/open-toast/protokt/pull/259/commits/0cd3157ccf84b8e79485d892bfdc1a7d748e8900 for what the implementation looks like given this PR

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -17,7 +17,7 @@ assertj = { module = "org.assertj:assertj-core", version.ref = "assertj" }
 buf = { module = "build.buf:buf", version.ref = "buf" }
 cel = { module = "org.projectnessie.cel:cel-bom", version.ref = "cel" }
 cel-core = { module = "org.projectnessie.cel:cel-core" }
-errorprone = { module = "com.google.errorprone:error_prone_core", version = "2.33.0" }
+errorprone = { module = "com.google.errorprone:error_prone_core", version = "2.34.0" }
 guava = { module = "com.google.guava:guava", version = "33.3.1-jre" }
 ipaddress = { module = "com.github.seancfoley:ipaddress", version.ref = "ipaddress" }
 jakarta-mail-api = { module = "jakarta.mail:jakarta.mail-api", version = "2.1.3" }


src/main/java/build/buf/protovalidate/internal/evaluator/EnumEvaluator.java
```diff
@@ -62,11 +62,8 @@ public boolean tautology() {
    */
   @Override
   public ValidationResult evaluate(Value val, boolean failFast) throws ExecutionException {
-    Descriptors.EnumValueDescriptor enumValue = val.value(Descriptors.EnumValueDescriptor.class);
-    if (enumValue == null) {
-      return ValidationResult.EMPTY;
-    }
-    if (!values.contains(enumValue.getNumber())) {
+    Integer enumValue = val.jvmValue(Integer.class);
+    if (!values.contains(enumValue)) {
```
pkwarren, line 66: I think we should keep the null check from before as `Set.contains` is defined that it could throw a NPE if the underlying set doesn't support null values.

src/main/java/build/buf/protovalidate/internal/evaluator/MessageLike.java
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buf.protovalidate.internal.evaluator;
+
+import com.google.protobuf.Descriptors.FieldDescriptor;
+
+public interface MessageLike {
```
pkwarren, line 19: How is this interface intended to be consumed by others? Should this be moved to another non-internal package?

src/main/java/build/buf/protovalidate/internal/evaluator/MessageLike.java
```diff
@@ -0,0 +1,23 @@
...
+  boolean hasField(FieldDescriptor field);
+
+  Value getField(FieldDescriptor field);
```
pkwarren, line 22: Please add Javadoc to the interface type and messages.

src/main/java/build/buf/protovalidate/internal/evaluator/MessageValue.java
```diff
@@ -18,33 +18,39 @@
...
+  public MessageLike messageValue() {
+    return value;
   }
 
   @Override
-  public <T> T value(Class<T> clazz) {
-    return clazz.cast(value);
+  @Nullable
+  public <T> T jvmValue(Class<T> clazz) {
+    return null;
```
pkwarren, line 48: Is this not used by anyone? Could we not keep it returning the underlying `Message` as before?

Do we need to make smaller interfaces and only implement `jvmValue` on `Value` types that provide a valid implementation?

src/main/java/build/buf/protovalidate/internal/evaluator/ObjectValue.java
```diff
@@ -50,17 +49,25 @@ public final class ObjectValue implements Value {
     this.value = value;
   }
 
-  @Nullable
   @Override
-  public Message messageValue() {
-    if (fieldDescriptor.getType() == Descriptors.FieldDescriptor.Type.MESSAGE) {
-      return (Message) value;
+  public MessageLike messageValue() {
+    return new ProtobufMessageLike((Message) value);
```
pkwarren, line 54: Why is this behavior changing to not return null if the underlying field isn't a message?

src/main/java/build/buf/protovalidate/internal/evaluator/ObjectValue.java
```diff
@@ -50,17 +49,25 @@ public final class ObjectValue implements Value {
...
+  }
+
+  @Override
+  public <T> T jvmValue(Class<T> clazz) {
+    if (value instanceof Descriptors.EnumValueDescriptor) {
+      return clazz.cast(((Descriptors.EnumValueDescriptor) value).getNumber());
+    } else {
```
pkwarren, line 61: nit: Remove the else and just return.

src/main/java/build/buf/protovalidate/internal/evaluator/EnumEvaluator.java
andrewparmet: Makes sense, I can do that. It's been a long time since wrote this - I don't remember if there was a reason I removed the null check or if I just tried it and it worked. Certainly the implementations used in this class now permit null, but it's good to safeguard.

src/main/java/build/buf/protovalidate/internal/evaluator/MessageLike.java
andrewparmet: This is the critical interface for other implementations to implement, so yes, it should move to a non-internal package. I didn't pick one since that tends to be something maintainers want to do! I'll move it now and you all can bike-shed from there.

src/main/java/build/buf/protovalidate/internal/evaluator/MessageValue.java
andrewparmet: Correct, this is not used. `jvmValue` is used for fields that protovalidate-java validates directly itself, like enums (as opposed to passing into CEL), and right now it does not do any validation on message types.

See also the changed doc on `Value.jvmValue`.

src/main/java/build/buf/protovalidate/internal/evaluator/ObjectValue.java
andrewparmet: I think what's happening is protovalidate always calls this correctly - only when the underlying field is a message. Just like the Set NPE you mention above, it's not a bad idea to defend against accidental bugs.

src/main/java/build/buf/protovalidate/internal/evaluator/MessageValue.java
andrewparmet: But even if it did, it doesn't make much sense to cast the message type to anything, does it? This hints at the narrower interface you mention.

PR #164, opened 2024-08-14
Editions fixes

jchadwick-buf:
Updates Protobuf to v27.3 and projectnessie/cel-java to 0.5.0. Fixes cases where delimited messages are handled improperly.

This is possible thanks to the work done in https://github.com/projectnessie/cel-java/pull/588.


pkwarren (commented):
We should also update src/test/resources/proto/buf.gen.*.yaml too.

jchadwick-buf:
Right, forgot about those ones. Done.

pkwarren (approved):

PR #202, opened 2024-10-27
Introduce MessageReflector to enable alternative protobuf runtimes

andrewparmet:
This PR implements a minimal abstraction in protovalidate-java to support additional protobuf runtimes (see #80).

It is a rehash of #132 since that PR ended up with some bad merge conflicts. See https://github.com/open-toast/protokt/pull/402 for what the implementation looks like given this PR.

The current integration in protokt works by converting protokt messages to `com.google.protobuf.DynamicMessage` upfront; the whole message tree converts on every `validate` call, regardless of which fields the validation rules actually read. I'd like to do less work.

To make this work end-to-end there are two halves:

1. **Teach protovalidate-java to understand protokt** for its internal validation (required-field checks, oneof resolution, field-level evaluator dispatch). This is the `MessageReflector`/`Value` API in this PR.
2. **Teach cel-java to understand protokt** for CEL expression evaluation (the `this` and `rules` variables inside rule expressions). This is where `celValue()` comes in and where further optimization lives.

Half #1 is clean and self-contained. Half #2 has a few possible directions with different trade-offs.

---

### `MessageReflector`

```java
public interface MessageReflector {
    boolean hasField(FieldDescriptor field);
    Value getField(FieldDescriptor field);
    Object celValue();
}
```

- `hasField` / `getField` drive protovalidate-java's non-CEL evaluators, such as required-field checks, oneof resolution, and Any's `type_url` lookup. Alternative runtimes implement these using their own reflection.
- `celValue()` is what CEL sees as the `this` variable when evaluating message-level rules. The implementation chooses whatever form CEL can navigate: a `com.google.protobuf.Message`, a `dev.cel.common.values.CelValue`, or anything else CEL recognizes. Part 2 is entirely about what to return here.

### `Value` (existing, made public)

```java
public interface Value {
    FieldDescriptor fieldDescriptor();
    MessageReflector messageValue();  // null if not a message
    List<Value> repeatedValue();
    Map<Value, Value> mapValue();
    Object celValue();                 // what CEL sees for this field
    <T> T jvmValue(Class<T> clazz);    // what protovalidate-java's own evaluators see
}
```

Alternative runtimes implement the MessageReflector and Value types. This falls out cleanly and matches how protovalidate-java already represents field values internally.

### `Validator` (gains a new overload)

```java
public interface Validator {
    // Existing:
    ValidationResult validate(Message msg) throws ValidationException;

    // New:
    ValidationResult validate(MessageReflector message, Descriptor descriptor)
        throws ValidationException;
}
```

### Status

All protovalidate-java conformance tests pass. This is sufficient to make protokt with protovalidate work: protokt's `celValue()` can return a DynamicMessage and CEL's existing `DescriptorMessageProvider` handles it. This is the eager-conversion path we'd like to avoid. Part 2 is about what `celValue()` can return instead.

---

### The current approach: return a DynamicMessage from `celValue()`

This is what this PR allows (combined with protokt PR 402). It works, and CEL navigates DynamicMessages fine. The cost is O(size of message tree) per `validate` call: the moment `celValue()` is called on a Message field, its full protokt message gets converted into a parallel tree of DynamicMessage builders, every field copied over, regardless of which fields the validation rules actually read.

For messages where rules touch a small subset of fields this is wasteful.

### A better approach: `StructValue`

`dev.cel.common.values.StructValue<F, V>` is cel-java's interface for "any struct-like thing CEL can navigate." If `celValue()` returns a StructValue, CEL calls `select(fieldName)` lazily, paying cost only for fields validation actually reads. Wrapping a protokt message this way is O(1); nested message fields become another StructValue on access, also O(1).

The protokt side is straightforward:

```kotlin
internal class ProtoktStructValue(
    private val message: Message?,
    private val descriptor: Descriptor,
    private val context: RuntimeContext,
) : StructValue<String, ProtoktStructValue>() {

    override fun value() = this
    override fun celType(): CelType = StructTypeReference.create(descriptor.fullName)

    override fun select(field: String): Any {
        val fd = findField(field) ?: throw IllegalArgumentException(...)
        return fieldValue(fd)
    }

    override fun find(field: String): Optional<Any> {
        if (message == null) return Optional.empty()
        val fd = findField(field) ?: return Optional.empty()
        return if (message.hasField(fd)) Optional.of(fieldValue(fd)) else Optional.empty()
    }

    // fieldValue / scalarCelValue / etc. return appropriate CEL-native representations:
    //   integer fields → Long (or UnsignedLong for uint32/64, fixed32/64)
    //   bytes → CelByteString
    //   nested messages → another ProtoktStructValue (or unwrapped WKT for WKTs)
}
```

Full implementation in [`ProtoktStructValue.kt`](https://github.com/andrewparmet/protokt/blob/protovalidate-lazy/protokt-protovalidate/src/main/kotlin/protokt/v1/buf/validate/ProtoktStructValue.kt).

Getting this to work revealed that cel-java's CelValue path is under-exercised. Three patches are needed. Which ones depend on which runtime.

---

### Part 2a: legacy runtime (`standardCelBuilder` / `CelRuntimeLegacyImpl`)

This is the default runtime and what protovalidate-java currently uses. To get StructValue semantics, `CelOptions.enableCelValue(true)` must be set. That trigger surfaces three latent bugs:

#### Patch 1: `FIXED32`/`FIXED64` handling in `ProtoCelValueConverter.fromProtoMessageFieldToCelValue`

**File**: `common/src/main/java/dev/cel/common/values/ProtoCelValueConverter.java`

The switch handles `UINT32` and `UINT64` explicitly, wrapping as `UnsignedLong`. It has no case for `FIXED32`/`FIXED64`; those fall through to `normalizePrimitive` which produces `Integer`/`Long`.

The CEL spec says `fixed32/fixed64 → uint` ([langdef.md](https://github.com/google/cel-spec/blob/master/doc/langdef.md)). cel-java's own checker and its `DescriptorMessageProvider` → `ProtoAdapter.fieldToValueConverter` both correctly map them to `UnsignedLong`. The CelValue path is the one outlier. When the checker says "this is uint64" but the runtime serves a `Long`, overload resolution fails: `No matching overload for function '_<=_'. Overload candidates: less_equals_uint64`.

**Archaeology**: the switch was introduced in commit `08347089` (2023-11-22, "Add a converter to handle bi-directional conversion between CelValue to Java native objects"). FIXED32/FIXED64 were never included. No commit since has ever touched the label, and there are no tests in cel-java that exercise fixed32/fixed64 against `ProtoCelValueConverter`; `ProtoMessageValueTest.java` only tests `single_uint32`.

**Fix**: add the missing case labels.

```java
case UINT32:
case FIXED32:
    if (!fieldDescriptor.isRepeated()) {
        return UnsignedLong.valueOf((int) result);
    }
    break;
case UINT64:
case FIXED64:
    if (!fieldDescriptor.isRepeated()) {
        return UnsignedLong.fromLongBits((long) result);
    }
    break;
```

**Risk**: low. Fixed32 and uint32 are wire-compatible unsigned types; treating them identically matches the CEL spec and the rest of cel-java.

**Test**: `ProtoMessageValueTest.selectField_fixed32_returnsUnsignedLong` and `selectField_fixed64_returnsUnsignedLong`. Branch: [`fix/fixed32-fixed64-handling`](https://github.com/andrewparmet/cel-java/tree/fix/fixed32-fixed64-handling).

#### Patch 2: preserve `FieldMask` as a message in the CEL runtime-value path

**File**: `common/src/main/java/dev/cel/common/values/ProtoCelValueConverter.java` (override).

`BaseProtoCelValueConverter.fromWellKnownProto` for `FIELD_MASK` converts the message to a comma-separated string of its `paths`. CEL expressions that access `fieldMask.paths` (e.g. protovalidate's `field_mask.const`, `field_mask.in`, `field_mask.not_in`) then fail with "Field selections must be performed on messages or maps."

The CEL spec's WKT conversion table lists `Any`, `ListValue`, `Struct`, `Value`, and the wrapper types as types that get unwrapped. It does not list `FieldMask` or `Empty`. Per spec, those should be treated as regular messages. For example, cel-go has zero mentions of `FieldMask` outside vendored protobuf types; it treats FieldMask as a normal message. cel-java's legacy-runtime `ProtoLiteAdapter.adaptValueToWellKnownProto` has an explicit comment: "These two WKTs are typically used in context of JSON conversions, in which they are automatically unwrapped into equivalent primitive types. In other cases, just return the original message itself." and returns the message as-is. Only cel-java's newer CelValue path unconditionally string-converts.

**Archaeology**: the `FIELD_MASK → string` conversion was added in commit `07376918` (2026-02-25, "Add conformance tests for planner"). Before that, `fromWellKnownProto` threw `UnsupportedOperationException` for `FIELD_MASK`. The commit's intent: make the `field_mask/to_json` case in `cel-spec/tests/simple/testdata/wrappers.textproto` pass on the planner:

```textproto
test {
  name: "to_json"
  container: "cel.expr.conformance.proto3"
  expr: "TestAllTypes{single_value: google.protobuf.FieldMask{paths: ['foo', 'bar']}}.single_value"
  value { string_value: "foo,bar" }
}
```

This test assigns a FieldMask to a `google.protobuf.Value` field. The commit put the string conversion in `fromWellKnownProto` so any `toRuntimeValue` of a FieldMask yields a string, which makes the JSON assignment work.

**Fix**: override `fromWellKnownProto` in `ProtoCelValueConverter` to wrap FieldMask as a `ProtoMessageValue` instead of string-converting.

```java
@Override
protected Object fromWellKnownProto(MessageLiteOrBuilder msg, WellKnownProto wellKnownProto) {
    switch (wellKnownProto) {
        case ANY_VALUE: /* existing */
        case FIELD_MASK:
            return ProtoMessageValue.create(
                (Message) msg, celDescriptorPool, this, celOptions.enableJsonFieldNames());
        default:
            return super.fromWellKnownProto((MessageOrBuilder) msg, wellKnownProto);
    }
}
```

**Concern about the `field_mask/to_json` test**: I initially worried this would regress that test, since it depends on `toRuntimeValue(FieldMask) == "foo,bar"`. It does not; the flow is:

1. `newValue(google.protobuf.FieldMask, {paths: [...]})` builds a FieldMask and calls `protoCelValueConverter.toRuntimeValue(FieldMask)`, which with the patch returns `ProtoMessageValue(FieldMask)`.
2. `EvalCreateStruct` (planner) and `CelValueRuntimeTypeProvider.createMessage` (legacy) both unwrap StructValue results via `.value()` / `maybeUnwrap` before returning. So the value passed to the outer `TestAllTypes{single_value: ...}` construction is the raw `FieldMask` message, not the ProtoMessageValue.
3. The outer construction calls `adaptValueToFieldType` → `adaptValueToProto(FieldMask, "google.protobuf.Value")` → `adaptValueToWellKnownProto(FieldMask, JSON_VALUE)` → `CelProtoJsonAdapter.adaptValueToJsonValue(FieldMask)`, which has a dedicated `FieldMask → "foo,bar"` branch at that level.

The string conversion lives in the JSON-assignment path, not in `fromWellKnownProto`. The commit that introduced the `fromWellKnownProto` branch was a convenience that also did the right thing for this test, but it isn't load-bearing. All of cel-java's own conformance tests (legacy and planner) pass with our override applied.

**Risk**: low on the JSON test. The concern flips the other direction: anywhere field access on FieldMask was previously failing silently because the runtime served a string, those paths now work. This is a behavior change.

**Test**: `ProtoMessageValueTest.selectField_fieldMask_returnsProtoMessageValue` sets a FieldMask with two paths, selects the field, asserts the result is a `ProtoMessageValue` whose `paths` field is a list of two strings. Without the patch, `select("field_mask")` returns `"foo,bar"` (string) and the assertion fails. Branch: [`fix/fieldmask-message-preservation`](https://github.com/andrewparmet/cel-java/tree/fix/fieldmask-message-preservation).

#### Patch 3: `CelValueRuntimeTypeProvider.newInstance` hierarchy regression

**File**: `runtime/src/main/java/dev/cel/runtime/CelValueRuntimeTypeProvider.java`

This one is legacy-only. The planner runtime uses a different code path that doesn't hit it.

When the legacy runtime has `enableCelValue(true)`, `CelRuntimeLegacyImpl` calls `CelValueRuntimeTypeProvider.newInstance(messageValueProvider)`. Inside, the code checks:

```java
if (valueProvider instanceof BaseProtoMessageValueProvider) {
    converter = ((BaseProtoMessageValueProvider) valueProvider).protoCelValueConverter();
} else if (valueProvider instanceof CombinedCelValueProvider) { /* ... */ }
```

The converter is used to turn raw protobuf Messages into SelectableValues for field access. Without a proper `ProtoCelValueConverter`, a default pass-through `BaseProtoCelValueConverter` is installed, and it can't wrap messages: 2000+ of 2872 conformance tests fail with "Field selections must be performed on messages or maps" on routine field access against the `rules` variable.

The problem: `ProtoMessageValueProvider` used to extend `BaseProtoMessageValueProvider`, so the `instanceof` check caught it. A later refactor stopped extending and instead implements `CelValueProvider` directly. The `instanceof` check in `newInstance` was not updated.

**Archaeology**:
- Commit `8d2c7e3d` (2025-04-29, "Add BaseProtoMessageValueProvider to decouple full/lite protobuf value providers") introduced the inheritance: `public class ProtoMessageValueProvider extends BaseProtoMessageValueProvider`. Method signature: `public BaseProtoCelValueConverter protoCelValueConverter()`.
- Commit `8d6bff68` (2026-02-02, "Introduce top-level runtime APIs based on Program Planner") removed the inheritance: `public class ProtoMessageValueProvider implements CelValueProvider`. Method renamed: `public CelValueConverter celValueConverter()` (different signature). `CelValueRuntimeTypeProvider.java` was not touched.

**Why nobody noticed**: the buggy path fires only for `enableCelValue(true)` on the legacy runtime. cel-java's default `standardCelBuilder` keeps `enableCelValue(false)`. cel-java's own test suite has three `enableCelValue(true)` hits, all in `CelLiteRuntime*Test`, which uses a different runtime entirely. And the planner runtime, which is more actively developed, uses a different code path.

**Fix**: extend the check to recognize `ProtoMessageValueProvider` directly via its new method.

```java
private static Optional<BaseProtoCelValueConverter> extractProtoCelValueConverter(
        CelValueProvider valueProvider) {
    if (valueProvider instanceof BaseProtoMessageValueProvider) {
        return Optional.of(((BaseProtoMessageValueProvider) valueProvider).protoCelValueConverter());
    }
    if (valueProvider instanceof ProtoMessageValueProvider) {
        CelValueConverter c = ((ProtoMessageValueProvider) valueProvider).celValueConverter();
        if (c instanceof BaseProtoCelValueConverter) {
            return Optional.of((BaseProtoCelValueConverter) c);
        }
    }
    if (valueProvider instanceof CombinedCelValueProvider) {
        return ((CombinedCelValueProvider) valueProvider).valueProviders().stream()
            .map(CelValueRuntimeTypeProvider::extractProtoCelValueConverter)
            .filter(Optional::isPresent).map(Optional::get).findFirst();
    }
    return Optional.empty();
}
```

**Risk**: low. Restores parity with the pre-refactor behavior. Callers passing a `BaseProtoMessageValueProvider` subclass keep working unchanged; callers passing a plain `ProtoMessageValueProvider` start working.

**Test**: `CelRuntimeLegacyImplTest.evaluate_enableCelValue_defaultProtoMessageValueProvider_fieldAccessSucceeds` builds a legacy `CelRuntime` with `enableCelValue(true)` and the default `ProtoMessageValueProvider`, evaluates `msg.single_int64` against a bound `TestAllTypes` variable, asserts the result is `42L`. Without the patch, the eval throws `CelEvaluationException: Error resolving field 'single_int64'. Field selections must be performed on messages or maps.` Branch: [`fix/protomessagevalueprovider-hierarchy`](https://github.com/andrewparmet/cel-java/tree/fix/protomessagevalueprovider-hierarchy).

---

### Part 2b: planner runtime (`plannerCelBuilder` / `CelRuntimeImpl`)

The planner runtime ignores `enableCelValue` and always walks CelValues internally. So the same protokt code (`ProtoktStructValue`) works on the planner too, and we need only Patches 1 and 2.

Switching protovalidate-java to the planner requires some changes. The planner bundle builder (`plannerCelBuilder`) throws on `setStandardEnvironmentEnabled(false)`, which protovalidate-java uses to exclude the `matches` function (replaced with a caching version in `CustomOverload` with a note to [google/cel-java#1038](https://github.com/google/cel-java/issues/1038)). The workaround is to construct the compiler and runtime separately and combine them:

```java
CelCompiler compiler =
    CelCompilerImpl.newBuilder(
            CelParserImpl.newBuilder(),
            CelCheckerLegacyImpl.newBuilder().setStandardEnvironmentEnabled(false))
        .setOptions(CEL_OPTIONS)
        .setStandardDeclarations(
            CelStandardDeclarations.newBuilder()
                .excludeFunctions(CelStandardDeclarations.StandardFunction.MATCHES)
                .build())
        .addLibraries(validateLibrary, CelExtensions.strings())
        .build();
CelRuntime runtime =
    CelRuntimeImpl.newBuilder()
        .setOptions(CEL_OPTIONS)
        .setStandardFunctions(
            CelStandardFunctions.newBuilder()
                .excludeFunctions(CelStandardFunctions.StandardFunction.MATCHES)
                .build())
        .addLibraries(validateLibrary, CelExtensions.strings())
        .build();
return CelFactory.combine(compiler, runtime);
```

The planner also requires `enableHeterogeneousNumericComparisons(true)` in `CelOptions`. Otherwise `CelOptions.current()` defaults are fine; `enableCelValue` stays off since the planner walks CelValues internally regardless.

If cel-java fixed the `matches` regex cache issue upstream, protovalidate-java could drop the `CustomOverload` caching replacement and its `setStandardEnvironmentEnabled(false)` / `excludeFunctions(MATCHES)` configuration entirely, at which point the planner scenarios just call `CelFactory.plannerCelBuilder()` directly and need no manual compiler/runtime wiring.

#### Silent-fallthrough hazard

There's a sharp edge in the legacy runtime's provider composition worth calling out. Nothing in protovalidate-java currently triggers it, but the behavior is surprising enough that anyone later adding a custom `CelValueProvider` to protovalidate-java should be aware of it.

If protovalidate-java uses the legacy runtime with `enableCelValue(true)` and passes a custom `CelValueProvider`, that provider replaces the default `ProtoMessageValueProvider` rather than composing with it:

```java
// CelRuntimeLegacyImpl.build()
CelValueProvider messageValueProvider = celValueProvider;
if (messageValueProvider == null) {
    messageValueProvider = ProtoMessageValueProvider.newInstance(options, dynamicProto);
}
runtimeTypeProvider = CelValueRuntimeTypeProvider.newInstance(messageValueProvider);
```

Any proto message types the custom provider doesn't know about fall through to the default pass-through converter, which can't wrap them, resulting in "Field selections must be performed on messages or maps."

The planner composes differently:

```java
// CelRuntimeImpl.build()
if (valueProvider() != null) {
    protoMessageValueProvider =
        CombinedCelValueProvider.combine(protoMessageValueProvider, valueProvider());
}
```

The planner wraps a user-supplied provider together with `ProtoMessageValueProvider`. A custom provider can return `Optional.empty()` for types it doesn't handle, falling through to proto-message navigation.

The prototyped integration doesn't set a custom `CelValueProvider` and just returns StructValue from `celValue()`. The CEL runtime's default `ProtoMessageValueProvider` handles field access on protobuf messages, and it passes our StructValue through unchanged because `CelValueConverter.toRuntimeValue` has `if (value instanceof CelValue) return value;`, so we never hit this.

The hazard only materializes if protovalidate-java later adds a custom provider (e.g. to teach CEL about some proto extension). On the legacy runtime, that would silently drop the default `ProtoMessageValueProvider`, and the protokt integration would break with the error surfacing on the `rules` variable itself.

The planner doesn't have this behavior.

---

## Implementation Options

I have three working combinations with different trade-offs. All three pass the full conformance suites.

### Option 1: legacy runtime, DynamicMessage

The current baseline demonstrated by this PR. protokt's `celValue()` returns a DynamicMessage. CEL's `DescriptorMessageProvider` navigates it, with eager conversion cost on every `validate` call.

**Branches required:**
- protovalidate-java: this PR
- protokt: [`protovalidate-lazy`](https://github.com/andrewparmet/protokt/tree/protovalidate-lazy) commit 1 only (*Use new protovalidate-java MessageReflector public API*)
- cel-java: none

**Pros:** zero patches to cel-java. Smallest surface area. Works with any CEL version.

**Cons:** eager conversion, which is the cost we want to avoid.

### Option 2: legacy runtime, StructValue

Use the legacy runtime, but protokt's `celValue()` returns `ProtoktStructValue` instead of DynamicMessage. Requires `enableCelValue(true)` on the CEL options. All three cel-java patches are needed to make it work.

**Branches required:**
- protovalidate-java: [`structvalue-experiment`](https://github.com/andrewparmet/protovalidate-java/tree/structvalue-experiment): sets `enableCelValue(true)` on top of the MessageReflector API
- protokt: [`protovalidate-lazy`](https://github.com/andrewparmet/protokt/tree/protovalidate-lazy) (both commits)
- cel-java: all three patch branches: [`fix/fixed32-fixed64-handling`](https://github.com/andrewparmet/cel-java/tree/fix/fixed32-fixed64-handling), [`fix/fieldmask-message-preservation`](https://github.com/andrewparmet/cel-java/tree/fix/fieldmask-message-preservation), [`fix/protomessagevalueprovider-hierarchy`](https://github.com/andrewparmet/cel-java/tree/fix/protomessagevalueprovider-hierarchy)

**Pros:** lazy field access. No DynamicMessage conversion per validate call.

**Cons:** depends on three cel-java patches. Two of them (Patches 1 and 2) are genuine latent bugs anyone hitting the CelValue path should want fixed; the third (Patch 3) is specific to the legacy runtime's provider-composition code.

### Option 3: planner runtime, StructValue

Same protokt code as Option 2, planner on protovalidate-java's side instead of the legacy runtime. Lazy field access, no DynamicMessage, and the planner's provider composition avoids the legacy-runtime sharp edge described above.

**Branches required:**
- protovalidate-java: [`planner-scenario-d`](https://github.com/andrewparmet/protovalidate-java/tree/planner-scenario-d): planner wiring only, no StructValue-specific CelOptions changes beyond what the planner itself requires.
- protokt: [`protovalidate-lazy`](https://github.com/andrewparmet/protokt/tree/protovalidate-lazy) (both commits)
- cel-java: Patches 1 and 2: [`fix/fixed32-fixed64-handling`](https://github.com/andrewparmet/cel-java/tree/fix/fixed32-fixed64-handling) and [`fix/fieldmask-message-preservation`](https://github.com/andrewparmet/cel-java/tree/fix/fieldmask-message-preservation). Patch 3 is not needed; the planner uses a different code path.

**Pros:** everything Option 2 gives plus the planner's composition behavior. Fewest cel-java patches among the StructValue options.

**Cons:** depends on the planner runtime, which is a larger protovalidate-java change than just enabling a CelOptions flag. Currently requires manual compiler/runtime wiring to preserve the `matches` workaround; collapses to a one-line `CelFactory.plannerCelBuilder()` if [google/cel-java#1038](https://github.com/google/cel-java/issues/1038) is fixed upstream and protovalidate-java's workaround can be dropped.

### How the options relate

Option 1 is where we are today. Option 2 is the most direct upgrade: keep the current runtime, just make `celValue()` smarter. Option 3 is the same shape as Option 2 but on the planner runtime, and exists as the long-term endpoint if protovalidate-java eventually moves to the planner for independent reasons.

### Aside: planner runtime + DynamicMessage

There's a fourth combination: planner runtime with protokt's `celValue()` still returning DynamicMessage (no StructValue). It's buildable ([`planner-scenario-b`](https://github.com/andrewparmet/protovalidate-java/tree/planner-scenario-b) demonstrates it passing conformance on Patches 1 and 2), but it's not a standalone option worth picking since it pays the cost of moving protovalidate-java to the planner without the StructValue benefit that motivates the move. It's a potential preview of what protovalidate-java's planner migration looks like in isolation.

---

## Branches

- **protovalidate-java** ([andrewparmet/protovalidate-java](https://github.com/andrewparmet/protovalidate-java)):
    - [`enable-lazy-protokt-evaluation`](https://github.com/andrewparmet/protovalidate-java/tree/enable-lazy-protokt-evaluation): MessageReflector/Value/validate(MessageReflector, Descriptor) API. Corresponds to PR [bufbuild/protovalidate-java#202](https://github.com/bufbuild/protovalidate-java/pull/202).
    - [`structvalue-experiment`](https://github.com/andrewparmet/protovalidate-java/tree/structvalue-experiment): adds `enableCelValue(true)`. For Option 2.
    - [`planner-scenario-d`](https://github.com/andrewparmet/protovalidate-java/tree/planner-scenario-d): planner wiring, `enableCelValue(true)`. For Option 3.
    - [`planner-scenario-b`](https://github.com/andrewparmet/protovalidate-java/tree/planner-scenario-b): planner wiring with `enableCelValue(false)`. Not a real option; see "Aside: planner runtime + DynamicMessage".
- **protokt** ([andrewparmet/protokt](https://github.com/andrewparmet/protokt) branch `protovalidate-lazy`, corresponds to PR [open-toast/protokt#402](https://github.com/open-toast/protokt/pull/402)). Two commits:
    1. *Use new protovalidate-java MessageReflector public API in protokt-protovalidate*: baseline integration. `celValue()` still returns DynamicMessage. For Option 1.
    2. *Return StructValue from MessageReflector.celValue() to avoid DynamicMessage*: the optimization. For Options 2 and 3.
- **cel-java** ([andrewparmet/cel-java](https://github.com/andrewparmet/cel-java)), one branch per patch:
    - [`fix/fixed32-fixed64-handling`](https://github.com/andrewparmet/cel-java/tree/fix/fixed32-fixed64-handling): Patch 1. Required for Options 2 and 3.
    - [`fix/fieldmask-message-preservation`](https://github.com/andrewparmet/cel-java/tree/fix/fieldmask-message-preservation): Patch 2. Required for Options 2 and 3.
    - [`fix/protomessagevalueprovider-hierarchy`](https://github.com/andrewparmet/cel-java/tree/fix/protomessagevalueprovider-hierarchy): Patch 3. Required for Option 2 only.


andrewparmet:
I'll rebase, no problem.

As far as a test case, I had implemented the validation in protokt's repository itself. Would you like me to move that implementation here? Or pull in the protokt runtime as a test dependency? 

I'm open to whatever strategy you all want here but don't want to do the scaffolding work if you decide to structure it differently. For now, this PR just includes what we'd need to implement protovalidate within protokt (which does support it right now, but only through conversion to DynamicMessage).

andrewparmet:
@srikrsna-buf Okay, I've updated the link to the protokt PR, which contains instructions to run the conformance tests there (after publishing this branch to Maven local). The new implementation still fails some conformance tests but I'm slowly bringing that number down.

andrewparmet:
@chrispine yes, thank you! Open to alternative API suggestions as well - this is just one way to get interoperation with other runtimes to work.

src/main/java/build/buf/protovalidate/MessageValue.java
```diff
@@ -42,13 +42,18 @@ final class MessageValue implements Value {
...
+  public MessageReflector messageValue() {
+    return value;
   }
 
   @Override
-  public <T> T value(Class<T> clazz) {
-    return clazz.cast(value);
+  public Object celValue() {
+      return value.getMessage();
+  }
```
srikrsna-buf, line 52: The CEL library needs to support the protobuf runtime, otherwise this will never be recognized as the protobuf message.

andrewparmet: I'm not sure I understand what you mean here - when the protovalidate runtime (this project) requests the value of this MessageValue object using this method, it's intending to pass the value into the CEL runtime. 

Looking at the usage today in CelPrograms:

```java
CelVariableResolver bindings = Variable.newThisVariable(val.value(Object.class));
```

where the return value of `value(Object.class)` is just `this.value`, which is a protobuf `Message`.

After this PR, this becomes:

```java
CelVariableResolver bindings = Variable.newThisVariable(val.celValue());
```

which is a more specific request than "cast this to an Object" - it's "please give me this value as it can be interpreted by CEL." The answer happens to be the same, as the new implementation returns `value.getMessage()`, which is the same protobuf `Message` value as before.

In the specific case of protokt, I have not implemented custom support for CEL to read protokt's `protokt.v1.Message` format, so when this is requested from a protokt message we'll wrap the message in a DynamicMessage that CEL can read. It's an area we can optimize further later by providing reflection support for CEL.

andrewparmet: Right now, when asked to evaluate a Message for CEL, the protokt runtime converts it to a protobuf-java DynamicMessage. This is the same thing it does today for the _root_ message object to enable protovalidate today, but is suboptimal since we can access basic fields (i.e., those used in validations evaluated from this library instead of within CEL) without converting the entire message to a DynamicMessage.

The plan was: If we benchmark and find that full message conversion is common and a bottleneck, we'll build native CEL support for protokt's Message interface as an extension in CEL.

andrewparmet: I have updated the description with my investigation into supporting alternate runtimes directly.

PR #212, opened 2024-11-26
Strict conformance fixes

jchadwick-buf:
As it turns out, not specifying `--strict` to the conformance test runner will cause `--strict_message` and `--strict_error` to do nothing. The conformance runner will just silently treat any validation errors as matching. Yikes.

Thankfully, protovalidate-java only had a few issues hidden by this footgun, and they are easily rectified after a bit of debugging.

Most of these issues are really quite minor and mainly just have to do with consistency across languages, but notably we were not marking validation errors from map keys properly, so that will be worth rolling into a release as soon as possible.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -264,6 +264,7 @@ mavenPublishing {
 dependencies {
     annotationProcessor(libs.nullaway)
     api(libs.protobuf.java)
+    implementation(libs.protobuf.java.util)
     implementation(enforcedPlatform(libs.cel))
     implementation(libs.cel.core)
     implementation(libs.guava)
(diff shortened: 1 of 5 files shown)


build.gradle.kts
```diff
@@ -264,6 +264,7 @@ mavenPublishing {
 dependencies {
     annotationProcessor(libs.nullaway)
     api(libs.protobuf.java)
+    implementation(libs.protobuf.java.util)
```
pkwarren, line 267: This dependency pulls in several other transitive dependencies (like gson) - if possible could we write our own version of `Timestamps.toString()` for use in the library to not pull in new dependencies?

We should look to write it using `java.time.*` APIs which are immutable.

jchadwick-buf: Done, now it uses the `Instant`/`DateTimeFormatter` API instead.

pkwarren (approved):

PR #215, opened 2024-12-04
Add field and rule value to violations

jchadwick-buf:
Adds the ability to access the captured rule and field value from a Violation.

There's a bit of refactoring toil, so I tried to split the commits cleanly to make this easier to review.
- The first commit adds the `Violation` wrapper class and plumbs it through all uses of `Violation`.
- The second commit makes `Value` non-internal so we can use it to expose protobuf values in a cleaner fashion.
- The third commit actually implements filling the `fieldValue` and `ruleValue`fields.

**This is a breaking change.** The API changes in the following ways:
- `ValidationResult` now provides a new wrapper `Violation` type instead of the `buf.validate.Violation` message. This new wrapper has a `getProto()` method to return a `buf.validate.Violation` message, and `ValidationResult` now has a `toProto()` method to return a `buf.validate.Violations` message.


src/main/java/build/buf/protovalidate/Violation.java
```diff
@@ -0,0 +1,131 @@
...
+    @Nullable Value ruleValue;
+
+    /**
+     * Sets the underlying protobuf message that corresponds to the violation.
+     *
+     * @param proto The value to set for this field.
+     * @return The builder.
+     */
+    public Builder setProto(build.buf.validate.Violation proto) {
+      this.proto = proto;
```
jhump, line 42: This would probably be the right place to call `Objects.requireNonNull`, instead of delaying to `build()`. But, also, I would move this to being a required argument to `newBuilder`, since this is a required field.

src/main/java/build/buf/protovalidate/Violation.java
```diff
@@ -0,0 +1,131 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * {@link Violation} contains all of the collected information about an individual constraint
+ * violation.
+ */
+public class Violation {
+  private final build.buf.validate.Violation proto;
+  @Nullable Value fieldValue;
+  @Nullable Value ruleValue;
```
jhump, line 27: `private final` for both of these

src/main/java/build/buf/protovalidate/Violation.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /** Builds a Violation instance. */
+  public static class Builder {
+    @Nullable private build.buf.validate.Violation proto;
+    @Nullable Value fieldValue;
+    @Nullable Value ruleValue;
```
jhump, line 33: `private`

src/main/java/build/buf/protovalidate/Violation.java
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Constructs a new empty builder.
+   *
+   * @return A new empty builder instance.
+   */
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  Violation(
```
jhump, line 89: This should be private. (I know you'll get a synthetic package-private constructor in byte-code, in order for it to be called from the builder class. But that is almost never really an issue.)

src/main/java/build/buf/protovalidate/Violation.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Constructs a {@link Builder} with all fields set to match this {@link Violation}.
+   *
+   * @return A new {@link Builder} instance.
+   */
+  public Builder toBuilder() {
+    return new Builder().setProto(proto).setFieldValue(fieldValue).setRuleValue(ruleValue);
+  }
```
jhump, line 130: Is this useful? Seems unnecessary and unlikely to be used.

jhump: Never mind. I see this was to make it easier to insert the new non-proto into existing code that already made use of `toBuilder` all over the place, to revise/mutate the protos as the result was popped-up the stack.

That is rather unfortunate. In addition to meaning extra public API that we don't expect users to actually find useful, it also means potentially _lots_ of re-building new objects. It probably should have popped builders all the way up the stack so it could just mutate and then `build()` only at the end when returning the value to the caller... Though a separate internal accumulator type would likely have been even better.

I suspect it might be a pain, but maybe worth tackling anyway? (Maybe not in this PR though, but in a follow-up, before we release this stuff, to minimize breaking changes?)

src/main/java/build/buf/protovalidate/Value.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package build.buf.protovalidate.internal.evaluator;
```
jhump, line 15: Unclear if this type really carries its weight as a top-level class in the public API. Maybe the `Violation` wrapper type could just return `Object` for field and rule values, and the Java docs could describe all of the possible types it could return? I assume that would be limited to the three composite types, `Message` (and concrete sub-types), `List<Value>`, and `Map<Value,Value>`, plus `String`, `byte[]`, and boxed primitives for the numeric/bool types?

src/main/java/build/buf/protovalidate/Violation.java
```diff
@@ -0,0 +1,131 @@
...
+    this.fieldValue = fieldValue;
+    this.ruleValue = ruleValue;
+  }
+
+  /**
+   * Gets the protobuf data that corresponds to this constraint violation.
+   *
+   * @return The protobuf violation data.
+   */
+  public build.buf.validate.Violation getProto() {
```
jhump, line 101: Let's call it `toProto`, for consistency?

src/main/java/build/buf/protovalidate/Violation.java
jchadwick-buf: What I really wanted to do was only expose an interface for the `Violation` and keep the actual implementation internal, but it felt like there was a lot to untangle to get it there.

That said, I like the idea of having an internal accumulator type. If I do make that separation throughout the entire codebase, then it's probably not even necessary to expose a concrete `Violation` class since nothing in protovalidate is going to accept a `Violation` object anyway. Or is that a bad idea?

src/main/java/build/buf/protovalidate/Value.java
jchadwick-buf: I think if we're going to do that, we should probably directly expose the `FieldDescriptor` like we do in Go. It's not technically needed, but I'm pretty sure it would be useful, and we always have it.

src/main/java/build/buf/protovalidate/Violation.java
jchadwick-buf: Done.

src/main/java/build/buf/protovalidate/Violation.java
jchadwick-buf: Done.

src/main/java/build/buf/protovalidate/Violation.java
jchadwick-buf: Done.

src/main/java/build/buf/protovalidate/Violation.java
jchadwick-buf: Done.

src/main/java/build/buf/protovalidate/Violation.java
jchadwick-buf: Done.

src/main/java/build/buf/protovalidate/Value.java
jchadwick-buf: Dropped Value from the public API + added corresponding FieldDescriptors to Violation. That said, I don't know what Java protobuf does for all of the dynamic cases, so I haven't actually added docs describing what the objects can be.

jchadwick-buf:
I'm going to try to pick up the violation accumulator change today if possible. Please feel free to review what's here though. Particularly interested if there's any way we can clean up the Violation API surface to be a bit nicer.

jchadwick-buf:
I did a bit of refactorings:
- Now `Violation` is an interface. There's an internal `ConstraintViolation` that implements it, and we stay in `Builder`s until the end.
- `Value` is still private, but there's a `Violation.FieldValue` interface to encapsulate the `Object` and `FieldDescriptor` for the field and rule values. This seems a bit less awkward.
- `Wrapper` interfaces, introduced just recently for structured field paths, are removed. Instead, I just mirrored what I started doing in the pv-go PR instead: carrying the necessary data on the `Value`. A bit tricky, but far simpler.
This necessitated some reshuffling that may not be intuitive and I'm sure there's ugly things in here but I wanted to get this out in some shape before the weekend.

src/main/java/build/buf/protovalidate/internal/evaluator/EmbeddedMessageEvaluator.java
```diff
@@ -0,0 +1,42 @@
...
+  @Override
+  public boolean tautology() {
+    return messageEvaluator.tautology();
+  }
+
+  @Override
+  public List<ConstraintViolation.Builder> evaluate(Value val, boolean failFast)
+      throws ExecutionException {
+    return FieldPathUtils.updatePaths(
+        messageEvaluator.evaluate(val, failFast), getFieldPathElement(), new ArrayList<>());
```
jhump, line 40: Does the last argument need to be mutable? I think it's probably better to use `Collections.emptyList()`.

src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBase.java
```diff
@@ -0,0 +1,61 @@
...
+package build.buf.protovalidate.internal.evaluator;
+
+import build.buf.protovalidate.internal.errors.FieldPathUtils;
+import build.buf.validate.FieldPath;
+import build.buf.validate.FieldPathElement;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+class EvaluatorBase {
```
jhump, line 24: Instead of extending this in every evaluator type, I think it would be better to use composition and maybe call it a `ConstraintViolationHelper` -- since it's really just used to get values for building a violation.

There's a lot of places that now call `super(evaluator)` that IMO don't look intuitive. Also, this doesn't provide any base `Evaluator` method implementations so it is awkward as a base class.

src/main/java/build/buf/protovalidate/internal/evaluator/AnyEvaluator.java
```diff
@@ -74,45 +77,42 @@ class AnyEvaluator implements Evaluator {
...
-              .build();
+      ConstraintViolation.Builder violation =
+          ConstraintViolation.newBuilder()
+              .addAllRulePathElements(getRulePrefixElements())
+              .addAllRulePathElements(IN_RULE_PATH.getElementsList())
+              .addFirstFieldPathElement(getFieldPathElement())
+              .setConstraintId("any.in")
+              .setMessage("type URL must be in the allow list")
+              .setFieldValue(new ConstraintViolation.FieldValue(val))
+              .setRuleValue(new ConstraintViolation.FieldValue(this.inValue, IN_DESCRIPTOR));
```
jhump, line 97: It would be nice to try to wrap some of this up into a helper in `EvaluatorBase`, to de-boiler-plate the various calls that do this.

It looks like they all call `ConstraintViolation.newBuilder().addAllRulePathElements(getRulePrefixElements())`, but the other methods vary slightly. Many also call `.addFirstFieldPathElement(getFieldPathElement())`, but some do not. Anyhow, maybe not worth looking into since they all seem to vary a little -- moving the builder pattern steps into method arguments would certainly make call sites more concise, but potentially at the cost of readability, so maybe not worth doing more here...

src/main/java/build/buf/protovalidate/Validator.java
```diff
@@ -75,11 +77,15 @@ public ValidationResult validate(Message msg) throws ValidationException {
...
     Descriptor descriptor = msg.getDescriptorForType();
     Evaluator evaluator = evaluatorBuilder.load(descriptor);
-    ValidationResult result = evaluator.evaluate(new MessageValue(msg), failFast);
-    if (result.isSuccess()) {
-      return result;
+    List<ConstraintViolation.Builder> result = evaluator.evaluate(new MessageValue(msg), failFast);
+    if (result.isEmpty()) {
+      return ValidationResult.EMPTY;
+    }
+    List<Violation> violations = new ArrayList<>();
```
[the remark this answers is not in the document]
jchadwick-buf: Good catch, fixed.

src/main/java/build/buf/protovalidate/internal/errors/ConstraintViolation.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.Deque;
+import java.util.List;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * {@link ConstraintViolation} contains all of the collected information about an individual
+ * constraint violation.
+ */
+public class ConstraintViolation implements Violation {
```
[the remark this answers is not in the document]
jchadwick-buf: Yep, same feelings here. I don't love it exactly, but actually it seems like a lot of Java software winds up doing basically this sort of thing, since there's really not much flexibility w.r.t. visibility.

src/main/java/build/buf/protovalidate/internal/evaluator/EmbeddedMessageEvaluator.java
jchadwick-buf: Oops, definitely doesn't need to be mutable.

src/main/java/build/buf/protovalidate/internal/evaluator/EvaluatorBase.java
jchadwick-buf: Yeah, I was going to try to extract more common functionality but it didn't work out so it's a bit pointless now. Now it's `ConstraintViolationHelper` and not being used as a base class.

src/main/java/build/buf/protovalidate/internal/evaluator/AnyEvaluator.java
jchadwick-buf: I am leaning on the other direction (not having `EvaluatorBase`) instead, since I'm on the side of this not being worth it, at least for now. It does feel like there's likely something more elegant to pull out of this, but with all of the edge cases we have right now it's not currently apparent.

src/main/java/build/buf/protovalidate/internal/evaluator/AnyEvaluator.java
```diff
@@ -34,7 +34,8 @@
  * com.google.protobuf.Any}'s within an expression, breaking evaluation if the type is unknown at
  * runtime.
  */
-class AnyEvaluator extends EvaluatorBase implements Evaluator {
+class AnyEvaluator implements Evaluator {
+  private final ConstraintViolationHelper constraintViolationHelper;
```
jhump, line 38: nit: I would call the field `helper` (here and throughout) to make the call sites more readable. They are so verbose, it's a bit of a chore to read or scan the code with this long name.

jhump (approved):
LGTM. One tiny nit, but not blocking if you disagree with it. If you do make the change, feel free to merge w/out further review.

PR #226, opened 2025-01-09
Refactor the API for v1

jchadwick-buf:
It's a bit ugly but seems like our best option since Java doesn't have any concept of `internal` packages. This mirrors the Go protovalidate package structure changes we did in https://github.com/bufbuild/protovalidate-go/pull/173.

The primary goal of this PR is to reduce the surface area of incidental public API. We don't want to expose our internal logic since it is very likely to be refactored in incompatible ways, but many things in the `internal` package are currently publicly accessible.


jhump (approved):
Makes sense to me.

PR #231, opened 2025-01-13
Update proto generated code to match runtime version

pkwarren:


buf.gen.yaml
```diff
@@ -1,4 +1,4 @@
 version: v2
 plugins:
-  - remote: buf.build/protocolbuffers/java:v29.1
+  - remote: buf.build/protocolbuffers/java:v29.3
```
pkwarren, line 3: Unfortunately this is a manual process for now: https://github.com/bufbuild/protovalidate-java/pull/230/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR8-R13

rodaine (approved):

PR #246, opened 2025-03-17
Jrinehart/tcn 2782 protovalidate java custom rules cannot use repeated nested

jrinehart-buf:
`protovalidate-java` cannot use a CEL expression where `this` represents a repeated message. `EvaluatorBuilder`'s `processFieldExpression` is currently directly using `Decls.newObjectType()` to create an object type based on the name of the message type in the field instead of honoring whether or not the field is repeated and returning a `Decls.newListType()`.

`provalidate` [recently added](https://github.com/bufbuild/protovalidate/pull/329) conformance tests that illustrate this problem. 

To fix this, this PR:

1. Adds JUnit tests and `.proto` that demonstrate the issue.
2. Updates `EvaluatorBuilder`'s `processFieldExpression` to delegate the determination of the CEL type to `DescriptorMappings.getCELType()`, which uses the `FieldDescriptor` to determine the correct type to return. This method requires passing a `forItems` flag stating if the type returned should be the list (or map) type for a field or the type of the items in the collection. Using the [new `nestedRule` property](https://github.com/bufbuild/protovalidate-java/pull/215)'s state provides the value of this flag, [similar to how it's used in `processStandardConstraints`](https://github.com/bufbuild/protovalidate-java/blob/b3cd73121d04182dc989b76902e468f94b5f0b99/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java#L421) to pass the `forItems` flag to `constraintCache.compile()`.
3. Because there were now 7 places where `valueEvaluator.getNestedRule() == null` was used to determine behavior, I moved this comparison into a `hasNestedRule()` method on `ValueEvaluator`, using conformance state to check regression.

Conformance notes:

This PR passes conformance in `protovalidate-java`. 

It has been tested against the conformance suite at the current head of `protovalidate` (which now includes the standard library, so there are many new failures).

Before changes in this PR, there were 137 failures. After, there were 135:

```
BEFORE: FAIL (failed: 137, skipped: 0, passed: 2681, total: 2818)
AFTER : FAIL (failed: 135, skipped: 0, passed: 2683, total: 2818)
```

Running `field_expression/repeated/message` shows that the two conformance tests fixed were for the desired issue:

```
BEFORE: FAIL (failed: 2, skipped: 0, passed: 2, total: 4)
AFTER:  PASS (failed: 0, skipped: 0, passed: 4, total: 4)
```

### DIFF
--- a/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
+++ b/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
@@ -334,7 +334,6 @@ class EvaluatorBuilder {
         ValueEvaluator valueEvaluatorEval)
         throws CompilationException {
       List<Constraint> constraintsCelList = fieldConstraints.getCelList();
-      boolean forItems = valueEvaluatorEval.hasNestedRule();
       if (constraintsCelList.isEmpty()) {
         return;
       }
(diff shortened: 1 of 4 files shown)


jchadwick-buf (approved):
Oops, good catch. LGTM.

PR #250, opened 2025-03-27
Enhance `isEmail` validation

smaye81:
This enhances the current isEmail validation by using a consistent regex that will be used across protovalidate implementations.

### DIFF
--- a/src/main/java/build/buf/protovalidate/CustomOverload.java
+++ b/src/main/java/build/buf/protovalidate/CustomOverload.java
@@ -27,7 +27,6 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.projectnessie.cel.common.types.BoolT;
 import org.projectnessie.cel.common.types.Err;
@@ -529,9 +528,7 @@ final class CustomOverload {
    * @return {@code true} if the input string is a valid email address, {@code false} otherwise.
    */
   private static boolean validateEmail(String addr) {
-    Matcher m = EMAIL_REGEX.matcher(addr);
-
-    return m.matches();
+    return EMAIL_REGEX.matcher(addr).matches();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -20,15 +20,14 @@
 import com.google.common.primitives.Bytes;
 import inet.ipaddr.IPAddress;
 import inet.ipaddr.IPAddressString;
-import jakarta.mail.internet.AddressException;
```
pkwarren, line 23: Are we now able to remove this dependency as well?

smaye81: mos def. removed.

pkwarren (approved):

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -58,6 +57,11 @@ final class CustomOverload {
   private static final String OVERLOAD_IS_INF = "isInf";
   private static final String OVERLOAD_IS_HOST_AND_PORT = "isHostAndPort";
 
+  // See https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+  private static final Pattern EMAIL_REGEX =
+      Pattern.compile(
+          "^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$");
```
timostamm, line 63: This expression is "JavaScript- and Perl-compatible". Is this expression supported by `java.util.regex.Pattern`?

pkwarren: This looks like supported syntax to me - earlier versions of Java didn't have some features (like named capture groups) but later ones added that.

smaye81: Yes, I tested this on regex101. Plus it builds and passes conformance and unit tests.

timostamm (approved):

PR #256, opened 2025-04-01
Enhance IP validation

smaye81:
This beefs up the validation for the following:

IP address (v4 and v6)
IP prefix (v4 and v6)
Hostname / Port


smaye81:
@pkwarren reverted back to JDK 8.

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -527,66 +557,81 @@ private static Val uniqueList(Lister list) {
...
-    try {
-      address = InetAddresses.forString(addr);
-    } catch (Exception e) {
-      return false;
-    }
-    if (ver == 0L) {
-      return true;
+  private static boolean isIP(String addr, long ver) {
+    if (ver == 6L) {
+      return new Ipv6(addr).address();
```
pkwarren, line 630: This is not very idiomatic Java here - having to create an empty object then calling a method on it to determine if it is valid.

I think we should make these functions like `isIPv4` and `isIPv6` that minimize allocations and state if possible.

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
+   * 2001:db8:abcd:0012:ffff:ffff:ffff:ffff.
+   *
+   * <p>An address prefix may include a specific host address, for example
+   * "2001:db8:abcd:0012::1f/64". With strict = true, this is not permitted. The host portion must
+   * be all zeros, as in "2001:db8:abcd:0012::0/64".
+   *
+   * <p>The same principle applies to IPv4 addresses. "192.168.1.0/24" designates the first 24 bits
+   * of the 32-bit IPv4 as the network prefix.
+   */
+  static boolean isIPPrefix(String str, long version, boolean strict) {
```
pkwarren, line 675:
```suggestion
  private static boolean isIPPrefix(String str, long version, boolean strict) {
```

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
+      Ipv4 ip = new Ipv4(str);
+      return ip.addressPrefix() && (!strict || ip.isPrefixOnly());
+    } else if (version == 0L) {
+      return isIPPrefix(str, 6, strict) || isIPPrefix(str, 4, strict);
+    }
+    return false;
+  }
+}
+
+final class Ipv4 {
```
pkwarren, line 689: I think we should extract this to a separate file. If we leave it here we should make this a static class as it uses nothing from the outer class.

```suggestion
static final class Ipv4 {
```

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
-    IPAddressString str;
-    IPAddress addr;
+  boolean isPrefixOnly() {
+    int bits = this.getBits();
+
+    int mask = 0;
+    if (this.prefixLen == 32) {
+      mask = 0xffffffff;
+    } else {
+      mask = ~(0xffffffff >>> this.prefixLen) >>> 0;
```
pkwarren, line 730: What is the `>>> 0` for here? I think it can be removed.

If we're pulling in Guava code I'd feel much more comfortable using their utilities to parse IP addresses instead of rolling our owns if they pass our tests.

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
+        && this.take('/')
+        && this.prefixLength()
+        && this.index == this.str.length();
+  }
+
+  // Stores value in `prefixLen`
+  private boolean prefixLength() {
+    int start = this.index;
+
+    while (true) {
```
pkwarren, line 755: The condition on line 756 should be moved into the while here.

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
+      }
+
+      if (this.index - start > 2) {
+        // max prefix-length is 32 bits, so anything more than 2 digits is invalid
+        return false;
+      }
+    }
+
+    String str = this.str.substring(start, this.index);
+    if (str.length() == 0) {
```
pkwarren, line 767:
```suggestion
    if (str.isEmpty()) {
```

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
+        // max 32 bits
+        return false;
+      }
+
+      this.prefixLen = val;
+
+      return true;
+
+    } catch (NumberFormatException nfe) {
+      // Error converting to number
```
pkwarren, line 790: Can remove this.

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
+      // bad leading 0
+      return false;
+    }
+
     try {
-      str = new IPAddressString(prefix);
-      addr = str.toAddress();
-    } catch (Exception e) {
+      int val = Integer.parseInt(str);
+
```
pkwarren, line 779: Lots of extra whitespace in here that isn't improving readability - I'd remove the majority of this.

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -611,40 +656,583 @@ private static boolean validateURI(String val, boolean checkAbsolute) {
...
+  /**
+   * Reports whether the current position is a digit.
+   *
+   * <p>Method parses the rule:
+   *
+   * <p>DIGIT = %x30-39 ; 0-9
+   */
+  private boolean digit() {
+    char c = this.str.charAt(this.index);
+    if ('0' <= c && c <= '9') {
```
pkwarren, line 864: We've got lots of places in here that could benefit from a `static boolean isDigit(char c)` helper to simplify code.

src/main/java/build/buf/protovalidate/CustomOverload.java
smaye81: OK, this is where the nuance of trying to stay consistent comes into play. This allocation is done across impls, so should we refactor those also? cc @timostamm

src/main/java/build/buf/protovalidate/CustomOverload.java
smaye81: Which outer class do you mean? This seems OK to have in the same file since it's a package-private class? FWIW, this is also an example of consistency. Should the other impls move their classes to separate files too? I originally had them separated in Go, but we moved everything into `library.go` to keep it in one place.

build.gradle.kts
```diff
@@ -271,7 +271,6 @@ dependencies {
     implementation(enforcedPlatform(libs.cel))
     implementation(libs.cel.core)
     implementation(libs.guava)
```
smaye81, line 273: We actually can't remove Guava. It's used in a bunch of places after looking more in-depth.

smaye81:
@timostamm @pkwarren think i addressed most comments. lmk if I missed anything.

pkwarren (approved):

PR #257, opened 2025-04-03
Enhance URI validation

smaye81:
This ports the validation logic from protovalidate-go for validating URIs and URI references.


src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -62,19 +60,19 @@ final class CustomOverload {
    */
   static Overload[] create() {
     return new Overload[] {
-      format(),
-      unique(),
-      startsWith(),
-      endsWith(),
-      contains(),
+      celFormat(),
```
smaye81, line 63: Renamed all these to stay consistent. I can un-rename them if it's confusing.

src/main/java/build/buf/protovalidate/Uri.java
```diff
@@ -0,0 +1,959 @@
...
+          if ('0' <= c && c <= '9') {
+            return c - '0';
+          } else if ('a' <= c && c <= 'f') {
+            return c - 'a' + 10;
+          } else if ('A' <= c && c <= 'F') {
+            return c - 'A' + 10;
+          }
+
+          return 0;
+        };
```
timostamm, line 297: IMO a `private static unhex` is preferable over the lambda here.

src/main/java/build/buf/protovalidate/Uri.java
```diff
@@ -0,0 +1,959 @@
...
+    return false;
+  }
+
+  /**
+   * Reports whether string contains a valid IPv6addrz.
+   *
+   * Method parses the rule:
+   *
+   * <pre>IPv6addrz = IPv6address "%25" ZoneID
+   */
```
timostamm, line 462: The methods are direct implementations of the ABNF rules. The comment on the reference implementation is:

```
  // Parses the rule from RFC 6874:
  //
  //     IPv6addrz = IPv6address "%25" ZoneID
```

This provides two important bits of information:
- The rule we're implementing. 
- Where it's defined, if not implied by preceding comments.

We have added summarizing sentences in Go and Python to follow best practices, but do we need them here? It's not clear what `string` refers to, and "contains" is misleading.

So I suggest:

```suggestion
  /**
   * Parses the rule from RFC 6874:
   *
   * <pre>IPv6addrz = IPv6address "%25" ZoneID
   */
```

Similar for all other methods.

timostamm (commented):
There's a new .DS_Store file that should be removed. The CI failure seems to be just a formatting issue.

src/main/java/build/buf/protovalidate/Uri.java
smaye81: Fixed. Should I fix the one in Go also?

src/main/java/build/buf/protovalidate/Uri.java
timostamm: Agreed: https://github.com/bufbuild/protovalidate-python/pull/277

timostamm (approved):
> Renamed all these to stay consistent. I can un-rename them if it's confusing.

I'm not sure that names and organization are ideal, but I don't have an idea that's decisively better.

PR #260, opened 2025-04-09
Fix ipv6 validation corner cases

smaye81:
This ports the ipv6 fix from https://github.com/bufbuild/protovalidate-go/pull/215.

For context see the description on the above PR. The summary is that this fixes the validation for some corner cases of IPv6 address validation. Namely:

* Adds a check that an IPv6 address can't begin or end on a single colon.
* Adds a check to fail-fast on invalid hextets.


src/main/java/build/buf/protovalidate/Ipv6.java
```diff
@@ -264,34 +272,39 @@ private boolean dotted() {
...
+      // hex digits at the current position.
       return false;
     }
 
     if (str.length() > 4) {
       // too long
-      return false;
+      // this is an error condition, it means we found a string of more than
+      // four valid hex digits, which is invalid in ipv6 addresses.
+      throw new ParseException("invalid hex", this.index);
```
pkwarren, line 299: If we want to use exceptions for control flow here, I'd throw an `IllegalStateException` here over a `ParseException`. 
```suggestion
      throw new IllegalStateException("invalid hex");
```

PR #263, opened 2025-04-16
Remove checked in generated code

pkwarren:
Update codegen to generate to the build directory to minimize the code in this project and match other protovalidate projects.

Upgrade to Gradle 8.13.


timostamm (approved):
IMO generating to the build directory works well for java (and dotnet). 
Where else do we do this? C++?

smaye81 (approved):
Nice. Ran this locally and confirmed `make` works for me.

pkwarren:
> IMO generating to the build directory works well for java (and dotnet). Where else do we do this? C++?

Looks like C++ does. I think for Java this makes the most sense. I'd like to get protobuf upgrades fully automated via dependabot where we can ensure when protobuf-java is upgraded that we also regenerate code on the latest version. Right now that is a manual step: https://github.com/bufbuild/protovalidate-java/blob/68daefff89c84ce18f04d23f1049638d17ef2927/gradle/libs.versions.toml#L8-L12

PR #265, opened 2025-04-20
Remove guava dependency

pkwarren:


build.gradle.kts
```diff
@@ -257,10 +257,10 @@ mavenPublishing {
 
 dependencies {
     annotationProcessor(libs.nullaway)
+    api(libs.jspecify)
```
pkwarren, line 260: jspecify appears to be the newer long term replacement for javax annotations, so switched to it. Previously we pulled these annotations in through Guava's jsr305 dependency.

conformance/build.gradle.kts
```diff
@@ -10,6 +10,12 @@ plugins {
     alias(libs.plugins.osdetector)
 }
 
+// Conformance tests aren't bound by lowest common library version.
```
pkwarren, line 13: No reason the tests themselves couldn't target a later version of Java.

src/main/java/build/buf/protovalidate/CustomOverload.java
```diff
@@ -205,12 +204,34 @@ private static Overload celContains() {
...
             byte[] receiver = (byte[]) lhs.value();
             byte[] param = (byte[]) rhs.value();
-            return Types.boolOf(Bytes.indexOf(receiver, param) != -1);
+            return Types.boolOf(bytesContains(receiver, param));
           }
           return Err.noSuchOverload(lhs, OVERLOAD_CONTAINS, rhs);
         });
   }
 
+  static boolean bytesContains(byte[] arr, byte[] subArr) {
```
pkwarren, line 213: This is the main thing we got from Guava. Immutable types are great but for our usage we didn't require them anywhere. We could always (when upgrading to Java 11) switch to similar types on the List/Map methods.

src/test/java/build/buf/protovalidate/CustomOverloadTest.java
```diff
@@ -181,38 +151,32 @@ public void testIsIpPrefixUnsupported() {
...
-          .isEqualTo(testCase.getValue());
-    }
+    assertThat(evalToBool("'foo@example.com'.isEmail()")).isTrue();
+    assertThat(evalToBool("'<foo@example.com>'.isEmail()")).isFalse();
+    assertThat(evalToBool("'  foo@example.com'.isEmail()")).isFalse();
+    assertThat(evalToBool("'foo@example.com    '.isEmail()")).isFalse();
+  }
+
+  @Test
+  public void testBytesContains() {
```
pkwarren, line 167: We may want some more coverage here in protovalidate, however I wanted to confirm the override worked after swapping out Guava.

rodaine (approved):

PR #277, opened 2025-04-30
Fix some formatting issues

smaye81:
This fixes a few issues noticed from failing conformance tests:

* Quotes were being escaped in string values in lists
* Durations had the word `duration` added to their formatting (along with escaped quotes)
* Doubles were not formatting the same as CEL-Go. Namely, insignificant digits were not being removed (`1.0` should format as `1`, not `1.0`)

In addition, this adds some additional unit tests for doubles to ensure the string formatting of doubles is adhering to the same behavior that CEL-Go uses.


build.gradle.kts
```diff
@@ -207,6 +207,13 @@ allprojects {
     }
     tasks.withType<Test>().configureEach {
         useJUnitPlatform()
+        this.testLogging {
+            events("failed")
+            exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+            showExceptions = true
+            showCauses = true
+            showStackTraces = true
+        }
```
smaye81, line 216: Added this bc it's much easier to see failing tests in the terminal than having to navigate out to a browser link to see the failing tests.

src/test/java/build/buf/protovalidate/FormatTest.java
```diff
@@ -15,18 +15,107 @@
...
+    assertThatThrownBy(
+            () -> {
+              Format.format("first value: %s and  %s", val);
+            })
+        .isInstanceOf(ErrException.class)
+        .hasMessageContaining("format: not enough arguments");
+  }
+
+  @Test
+  void testDouble() {
```
smaye81, line 45: @jchadwick-buf this is what I meant by some nuance in testing. We could move these double tests to the conformance tests I guess, but not sure about things like testing the above thrown exception. Wdyt?

src/main/java/build/buf/protovalidate/Format.java
```diff
@@ -34,6 +34,7 @@
 final class Format {
   private static final char[] HEX_ARRAY = "0123456789ABCDEF".toCharArray();
   private static final char[] LOWER_HEX_ARRAY = "0123456789abcdef".toCharArray();
+  private static final DecimalFormat decimalFormat = new DecimalFormat("0.#########");
```
pkwarren, line 37: DecimalFormat instances aren't thread safe - we should use a ThreadLocal to cache these (if we expect they'll be used a lot), add synchronization around them, or create them on demand:

> Decimal formats are generally not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.

https://docs.oracle.com/javase/8/docs/api/java/text/DecimalFormat.html

src/main/java/build/buf/protovalidate/Format.java
```diff
@@ -139,52 +140,25 @@ private static String bytesToHex(byte[] bytes, char[] digits) {
...
-      DecimalFormat format = new DecimalFormat();
-      format.setMaximumFractionDigits(Integer.MAX_VALUE);
-      format.setMinimumFractionDigits(6);
-      builder.append(format.format(val.value()));
     } else if (type == TypeEnum.String) {
-      builder.append("\"").append(val.value().toString()).append("\"");
+      builder.append(val.value());
     } else if (type == TypeEnum.Bytes) {
-      formatBytes(builder, val);
+      builder.append(new String((byte[]) val.value(), StandardCharsets.UTF_8));
```
pkwarren, line 149: Mainly a question on CEL in general but can we always assume bytes can be converted to a valid UTF-8 string? Is there some fallback where it would display invalid UTF-8 strings as hex or something like that?

src/main/java/build/buf/protovalidate/Format.java
```diff
@@ -139,52 +140,25 @@ private static String bytesToHex(byte[] bytes, char[] digits) {
...
+    } else if (type == TypeEnum.Int || type == TypeEnum.Uint || type == TypeEnum.Double) {
```
pkwarren, line 150: Shouldn't we only need to format float/double as decimals? It feels like for int/uint we could just output the string value.

src/test/java/build/buf/protovalidate/FormatTest.java
jchadwick-buf: Personally, I'm not overly concerned about handling all of the error cases identically as long as the error cases do fail as-expected, so I don't see this as a huge problem for adding tests to conformance per-se.

src/main/java/build/buf/protovalidate/Format.java
smaye81: Our formatting is pretty inconsistent right now and we need to add more comprehensive tests to make sure it behaves the same. Right now, CEL-Go's built-in formatting function will throw a runtime error with invalid UTF-8, but our Java implementation will print placeholders �� when using `%s` and will just print hex digits when using `%x`. It's on the docket to unify all this behavior soon.

pkwarren (approved):

PR #283, opened 2025-05-09
Refactor validator creation to support various configurations

### LINKED ISSUE #51: [BUG] Config.disableLazy isn't functional as written
## Description
The current implementation of `Config.disableLazy` isn't functional. In addition to specifying `disableLazy = true`, a user also needs to provide the initial set of messages/descriptors to use to initialize the cache.

## Steps to Reproduce
Pass a `Config` to `Validator` with `disableLazy = true`. Attempt to validate a message - nothing will work because it assumes the supported messages are already cached.

## Expected Behavior
The `Config` object should support passing the initial set of messages or descriptors to use to seed the cache with `disableLazy = true`.

## Actual Behavior
The `disableLazy = true` config option isn't usable.

## Screenshots/Logs
n/a

## Environment

- **Operating System**: macOS
- **Version**: 14.0
- **Compiler/Toolchain**: Java 17
- **Protobuf Compiler & Version**: protoc 24.4
- **Protoc-gen-validate Version**: n/a
- **Protovalidate Version**: 0.1.6

## Possible Solution
Pass a list of messages/descriptors in the `Config` object to allow initializing the cache prior to use with `disableLazy = true`.

## Additional Context
n/a

smaye81:
Fixes #51 

This makes a series of changes to the way that a validator is created. Doing so enables more flexible ways to configure a validator and also provides a bit more safety when attempting to create a validator with invalid state. The impetus behind this change was to allow for the ability to specify a list of seed descriptors when building a validator and for the optional ability to disable lazy-loading of any descriptors not provided.

Previously, a validator could be created either with a default constructor (`new Validator()`) or via a `Config` object (`new Validator(config)`). However, the `Config` object had no way to provide a list of descriptors, it only had a `disableLazy` field. Simply adding a list of descriptors to the `Config` presented a few problems though:

* Since these were two unrelated fields in the config, users had a serious footgun to play with because it was possible to set `disableLazy` to true and not specify any descriptors. Doing so would return an error at validation time with a validation message about an unknown evaluator.
* Even if the above was tolerable, building validation rules from descriptors necessitates catching a `CompilationException`. This meant that if the seed descriptors were added to the config, all users creating a validator would now have to catch a checked exception even if they weren't using the seed descriptors.

The solution in this PR converts validator creation from using constructors into a builder pattern with the following changes:
* `disableLazy` and the new `descriptors` list have been moved from the config and are now parameters at instantiation time.
* There are two options for building once the builder is created:
  *  `build()`
  *  `buildWithDescriptors(List<Descriptor>, boolean disableLazy) throws CompilationException, InvalidStateException`

The overload that accepts seed descriptors is the only one that throws a `CompilationException`. If users choose to use `build` with no arguments, no checked exception needs to be handled. 

Note that the above also allows us to provide a helpful check at instantiation time if a user sets `disableLazy` to true but doesn't provide any descriptors. In this case, an `IllegalStateException` is thrown. It is still possible to provide an _incorrect_ list of descriptors, but that is impossible to catch at instantiation time (but hopefully is now a bit harder to do).

The new API now adds a `ValidatorFactory` class with static methods for creating a validator. In addition, the type returned is now an interface (still named `Validator`) which provides us with a bit more flexibility both for testing and for making non-breaking changes in the future.

Examples of creating a validator with no seed descriptors:

```java
// Create a Validator with a default config
Validator validator = ValidatorFactory.newBuilder().build();

// Create a Validator with a custom config
Config cfg = Config.newBuilder().setFailFast(true).build();
Validator validator = ValidatoryFactory.newBuilder().withConfig(cfg).build();
```

Creating a validator with seed descriptors is similar. You would still call `newBuilder`, but at build time, you call `buildWithDescriptors` and pass the descriptors and set `disableLazy` accordingly. Using this approach requires catching exceptions at build time:

```java
// Create a validator with seed descriptors
MyMessage msg = MyMessage.newBuilder().build();
List<Descriptor> seedDescriptors = new ArrayList<Descriptor>();
seedDescriptors.add(msg.getDescriptorForType());

try {
    Validator validator = ValidatorFactory.newBuilder().buildWithDescriptors(seedDescriptors, false);

    // Note you can also create a validator with a config
    // Validator validator = ValidatorFactory.newBuilder().withConfig(cfg).buildWithDescriptors(seedDescriptors, false);

    // Also note that the buildWithDescriptors function is what throws the exception so the validator 
    // builder can be created ahead of time with no need for catching an exception.
} catch (CompilationException ce) {
    // Exception pre-warming the cache with the seed descriptors
} catch (IllegalStateException ise) {
    // This is thrown if seedDescriptors is empty and disableLazy is true
}
```

Note: **THIS IS A BREAKING CHANGE**, but the hope is that this makes it easier and more intuitive for creating a validator while providing future-proof flexibility. To migrate, users should make the following changes:

**With no config**

```diff
import build.buf.protovalidate.Validator;
+ import build.buf.protovalidate.ValidatorFactory;

- Validator validator = new Validator();
+ Validator validator = ValidatorFactory.newBuilder().build();
```

**With config**

```diff
import build.buf.protovalidate.Validator;
+ import build.buf.protovalidate.ValidatorFactory;

Config cfg = Config.newBuilder().setFailFast(true).build();
- Validator validator = new Validator(cfg);
+ Validator validator = ValidatorFactory.newBuilder().withConfig(cfg).build();
```

In addition, the `loadMessages` and `loadDescriptors` methods from the validator have been removed. Users should instead use the `newBuilder(List<Descriptor> descriptors, boolean disableLazy)` approach to load descriptors ahead of time.

### DIFF
--- a/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
+++ b/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
@@ -51,6 +51,7 @@ class EvaluatorBuilder {
 
   private final Env env;
   private final boolean disableLazy;
+  private final List<Descriptor> descriptors;
   private final RuleCache rules;
 
   /**
@@ -61,8 +62,27 @@ class EvaluatorBuilder {
    */
   public EvaluatorBuilder(Env env, Config config) {
     this.env = env;
-    this.disableLazy = config.isDisableLazy();
+    this.descriptors = new ArrayList<Descriptor>();
+    this.disableLazy = false;
     this.rules = new RuleCache(env, config);
+  }
+
+  /**
+   * Constructs a new {@link EvaluatorBuilder}.
+   *
+   * @param env The CEL environment for evaluation.
+   * @param config The configuration to use for the evaluation.
+   */
+  public EvaluatorBuilder(Env env, Config config, List<Descriptor> descriptors, boolean disableLazy)
+      throws CompilationException {
+    this.env = env;
+    this.descriptors = descriptors;
+    this.disableLazy = disableLazy;
+    this.rules = new RuleCache(env, config);
+
+    for (Descriptor descriptor : this.descriptors) {
+      this.build(descriptor);
+    }
   }
 
   /**
--- a/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
+++ b/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
@@ -73,14 +73,13 @@ class EvaluatorBuilder {
    */
   EvaluatorBuilder(Env env, Config config, List<Descriptor> descriptors, boolean disableLazy)
       throws CompilationException {
+    Objects.requireNonNull(descriptors, "descriptors must not be null");
     this.env = env;
     this.disableLazy = disableLazy;
     this.rules = new RuleCache(env, config);
 
-    if (descriptors != null) {
-      for (Descriptor descriptor : descriptors) {
-        this.build(descriptor);
-      }
+    for (Descriptor descriptor : descriptors) {
+      this.build(descriptor);
     }
   }
 
--- a/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
+++ b/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
@@ -60,7 +60,7 @@ class EvaluatorBuilder {
    * @param env The CEL environment for evaluation.
    * @param config The configuration to use for the evaluation.
    */
-  public EvaluatorBuilder(Env env, Config config) {
+  EvaluatorBuilder(Env env, Config config) {
     this.env = env;
     this.descriptors = new ArrayList<Descriptor>();
     this.disableLazy = false;
@@ -73,7 +73,7 @@ class EvaluatorBuilder {
    * @param env The CEL environment for evaluation.
    * @param config The configuration to use for the evaluation.
    */
-  public EvaluatorBuilder(Env env, Config config, List<Descriptor> descriptors, boolean disableLazy)
+  EvaluatorBuilder(Env env, Config config, List<Descriptor> descriptors, boolean disableLazy)
       throws CompilationException {
     this.env = env;
     this.descriptors = descriptors;
@@ -93,7 +93,7 @@ class EvaluatorBuilder {
    * @return An evaluator for the descriptor type.
    * @throws CompilationException If an evaluator can't be created for the specified descriptor.
    */
-  public Evaluator load(Descriptor desc) throws CompilationException {
+  Evaluator load(Descriptor desc) throws CompilationException {
     Evaluator evaluator = evaluatorCache.get(desc);
     if (evaluator == null && disableLazy) {
       return new UnknownDescriptorEvaluator(desc);
--- a/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
+++ b/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
@@ -51,7 +51,6 @@ class EvaluatorBuilder {
 
   private final Env env;
   private final boolean disableLazy;
-  private final List<Descriptor> descriptors;
   private final RuleCache rules;
 
   /**
@@ -62,7 +61,6 @@ class EvaluatorBuilder {
    */
   EvaluatorBuilder(Env env, Config config) {
     this.env = env;
-    this.descriptors = new ArrayList<Descriptor>();
     this.disableLazy = false;
     this.rules = new RuleCache(env, config);
   }
@@ -76,11 +74,10 @@ class EvaluatorBuilder {
   EvaluatorBuilder(Env env, Config config, List<Descriptor> descriptors, boolean disableLazy)
       throws CompilationException {
     this.env = env;
-    this.descriptors = descriptors;
     this.disableLazy = disableLazy;
     this.rules = new RuleCache(env, config);
 
-    for (Descriptor descriptor : this.descriptors) {
+    for (Descriptor descriptor : descriptors) {
       this.build(descriptor);
     }
   }
--- a/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
+++ b/src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
@@ -77,8 +77,10 @@ class EvaluatorBuilder {
     this.disableLazy = disableLazy;
     this.rules = new RuleCache(env, config);
 
-    for (Descriptor descriptor : descriptors) {
-      this.build(descriptor);
+    if (descriptors != null) {
+      for (Descriptor descriptor : descriptors) {
+        this.build(descriptor);
+      }
     }
   }
 
--- a/src/main/java/build/buf/protovalidate/ValidatorFactory.java
+++ b/src/main/java/build/buf/protovalidate/ValidatorFactory.java
@@ -33,7 +33,7 @@ public final class ValidatorFactory {
   /** A builder class used for building a validator. */
   public static class ValidatorBuilder {
     /** The config object to use for instantiating a validator. */
-    @Nullable protected Config config;
+    @Nullable private Config config;
 
     /**
      * Create a validator with the given config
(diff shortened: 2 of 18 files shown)


src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
```diff
@@ -51,6 +51,7 @@ class EvaluatorBuilder {
 
   private final Env env;
   private final boolean disableLazy;
+  private final List<Descriptor> descriptors;
```
pkwarren, line 54: This only looks to be referenced in the constructor. We should be able to remove this field and just iterate over the constructor arg below.

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -0,0 +1,159 @@
...
+import org.jspecify.annotations.Nullable;
+
+/**
+ * ValidatorFactory is used to create a validator.
+ *
+ * <p>Validators can be created with an optional {@link Config} to customize behavior. They can also
+ * be created with a list of seed descriptors to warmup the validator cache ahead of time as well as
+ * an indicator to lazily-load any descriptors not provided into the cache.
+ */
+public class ValidatorFactory {
```
pkwarren, line 30:
```suggestion
public final class ValidatorFactory {
```

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -0,0 +1,159 @@
...
+  // Prevent instantiation
+  private ValidatorFactory() {}
+  ;
```
pkwarren, line 33:
```suggestion
```

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Build the validator, warming up the cache with any provided descriptors.
+     *
+     * @return A Validator instance
+     * @throws CompilationException If any of the given descriptors' validation rules fail
+     *     processing while warming up the cache.
+     * @throws IllegalStateException If disableLazy is set to true and no descriptors are passed.
+     */
+    public Validator build() throws CompilationException, IllegalStateException {
+      if (disableLazy && this.descriptors.size() == 0) {
```
pkwarren, line 57:
```suggestion
      if (disableLazy && this.descriptors.isEmpty()) {
```

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -0,0 +1,159 @@
...
+   *     makes the entire cache read-only, eliminating thread contention.
+   * @return A Validator builder
+   */
+  public static EagerBuilder newBuilder(List<Descriptor> descriptors, boolean disableLazy) {
+    return new EagerBuilder(descriptors, disableLazy);
+  }
+}
+
+/** An abstract class that all validator builders should extend. */
+abstract class BaseBuilder<T extends BaseBuilder<T>> {
```
pkwarren, line 142: This should be made static and lifted into the ValidatorFactory class or moved to a separate .java file.

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * @param descriptors the list of descriptors to warm up the cache.
+   * @param disableLazy whether to disable lazy loading of validation rules. When validation is
+   *     performed, a message's rules will be looked up in a cache. If they are not found, by
+   *     default they will be processed and lazily-loaded into the cache. Setting this to false will
+   *     not attempt to lazily-load descriptor information not found in the cache and essentially
+   *     makes the entire cache read-only, eliminating thread contention.
+   * @return A Validator builder
+   */
+  public static EagerBuilder newBuilder(List<Descriptor> descriptors, boolean disableLazy) {
```
pkwarren, line 136: Perhaps a simpler design would be to have a single builder class, with two build methods:
```java
public Validator build() {}

public Validator buildWithDescriptors(List<Descriptor> descriptors, boolean disableLazy) throws CompilationException, IllegalStateException;
```

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  /**
+   * A convenience function for creating a new validator with a default configuration.
+   *
+   * @return A Validator instance
+   */
+  public static Validator defaultInstance() {
+    return newBuilder().build();
```
pkwarren, line 113: I think we should leave this out - it isn't that much more to type to have someone just do `ValidatorFactory.newBuilder().build()`. We could always add it back if desired.

This method as written feels like it is returning the same instance each time (similar to the global validator in pv-go: https://pkg.go.dev/buf.build/go/protovalidate#Validator), not a new instance each time w/ the default config.

src/main/java/build/buf/protovalidate/ValidatorFactory.java
smaye81: That's fair. To do that, we'd have to create the `EvaluatorBuilder` in the builder and keep it in state so that it can be passed to the `ValidatorImpl` constructor. Not sure if that's desirable or not. It definitely simplifies the design but at the cost of having more logic in the actual builder itself. Wdyt?

src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
```diff
@@ -76,12 +74,13 @@ class EvaluatorBuilder {
   EvaluatorBuilder(Env env, Config config, List<Descriptor> descriptors, boolean disableLazy)
       throws CompilationException {
     this.env = env;
-    this.descriptors = descriptors;
     this.disableLazy = disableLazy;
     this.rules = new RuleCache(env, config);
 
-    for (Descriptor descriptor : this.descriptors) {
-      this.build(descriptor);
+    if (descriptors != null) {
```
pkwarren, line 80: The nil check isn't needed here - if anything we could add an `Objects.requireNonNull(descriptors)` above if we wanted with an error message. Doesn't hurt but typically you never pass around null collections.

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -27,63 +26,28 @@
...
-    private final boolean disableLazy;
-
-    EagerBuilder(List<Descriptor> descriptors, boolean disableLazy) {
-      this.descriptors = Collections.unmodifiableList(descriptors);
-      this.disableLazy = disableLazy;
-    }
+  /** A builder class used for building a validator. */
+  public static class ValidatorBuilder {
+    /** The config object to use for instantiating a validator. */
+    @Nullable protected Config config;
```
pkwarren, line 36:
```suggestion
    @Nullable private Config config;
```

src/main/java/build/buf/protovalidate/ValidatorFactory.java
```diff
@@ -27,63 +26,28 @@
...
     /**
-     * Build the validator, warming up the cache with any provided descriptors.
+     * Create a validator with the given config
      *
-     * @return A Validator instance
-     * @throws CompilationException If any of the given descriptors' validation rules fail
-     *     processing while warming up the cache.
-     * @throws IllegalStateException If disableLazy is set to true and no descriptors are passed.
+     * @param config The {@link Config} to configure the validator.
+     * @return T the builder instance
```
pkwarren, line 42:
```suggestion
     * @return The builder instance
```

pkwarren (approved):

src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
smaye81: Without the check, it throws an NPE though:  `java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()"`. I think we should do _something_ here to catch that. Seems like an easy thing to prevent.

pkwarren: This would only happen though if someone passes `null` to the `buildDescriptors` method, right?

Typically if you wanted a nicer error you'd do `Objects.requireNonNull(descriptors, "descriptors must not be null")` as the first line in the constructor (or the equivalent Preconditions.checkNotNull from Guava).

pkwarren (approved):

PR #293, opened 2025-05-28
Add conformance testing to string.format implementation

smaye81:
Apologies for the large PR. The logical way to split this up is for `Format.java` to be its own PR with everything else separate but it may just be easier to review this all at once. So, an easy way to review is:
* `Format.java` for all the formatting fixes
*  Everything else that runs the conformance tests

This adds conformance testing to our implementation of `string.format`. cel-java does not currently provide a `string.format` function so we implement our own and add it to the protovalidate-java CEL runtime.

While formatting is not (yet) a part of the cel-spec, they do provide conformance test data for validating behavior. So, to ensure we are conformant, we use this test data and run it against our implementation.

See the test data [here](https://github.com/google/cel-spec/blob/master/tests/simple/testdata/string_ext.textproto). Note that this file contains various sections for testing. However, we are only concerned with the `format` and `format_errors` sections. In addition, since we are only testing for formatting, we do not handle every `oneof` case or variation in the `SimpleTest` definition -- only expected results or evaluation errors. If others are added down the road in future versions of the spec, we can adjust these tests.

There are few tests marked as skipped in `FormatTest` as these seem to be related to issues in cel-java's runtime (or in how we set up our environment) because the same issues happen when trying to use those expressions in protovalidate-java. Probably worth investigating or filing an issue for cel-java, but didn't want to hold up this PR.


src/main/java/build/buf/protovalidate/Format.java
```diff
@@ -93,153 +92,275 @@ static String format(String fmtString, ListT list) {
...
+        return val.value().toString();
+      case Int:
+      case Uint:
+        Optional<String> str = validateNumber(val);
+        if (str.isPresent()) {
+          return str.get();
+        }
+        return val.value().toString();
+      case Bytes:
+        return new String((byte[]) val.value(), StandardCharsets.UTF_8);
```
pkwarren, line 147: What does this do if the value is invalid UTF-8 here?

src/main/java/build/buf/protovalidate/Format.java
```diff
@@ -93,153 +92,275 @@ static String format(String fmtString, ListT list) {
...
-      Object obj = list.get(i);
-      formatString(builder, DefaultTypeAdapter.nativeToValue(Db.newDb(), null, obj));
-      if (i != list.size() - 1) {
+
+    IteratorT iter = val.iterator();
+    int index = 0;
+    while (iter.hasNext().booleanValue()) {
+      Val v = iter.next();
+      builder.append(formatString(v));
+      if (index != val.size().intValue() - 1) {
```
pkwarren, line 185: nit: Could use `iter.hasNext()` here instead of keeping track of an index.

src/test/java/build/buf/protovalidate/FormatTest.java
```diff
@@ -15,107 +15,210 @@
...
+          sections.stream()
+              .filter(s -> s.getName().equals("format_errors"))
+              .findFirst()
+              .map(SimpleTestSection::getTestList)
+              .orElse(Collections.emptyList());
+
+      env = Env.newEnv(Library.Lib(new ValidateLibrary()));
+
+    } catch (InvalidPathException | IOException e) {
+      fail(e);
```
pkwarren, line 118: I'd just add a `throws Exception` to the `setUp` method which will automatically fail the test.

pkwarren (approved):

src/main/java/build/buf/protovalidate/Format.java
smaye81: Talked offline, but in the interest of visibility:

Currently this will print a placeholder �, but the [strings.md](https://github.com/google/cel-spec/blob/master/doc/extensions/strings.md) doc on the cel-spec site specifies:

> The value is formatted as if string(value) was performed and any invalid UTF-8 sequences are replaced with \ufffd. Multiple adjacent invalid UTF-8 sequences must be replaced with a single \ufffd.

So, this is currently non-conformant, but there are no actual conformance tests yet to cover this. Will implement this in a follow-up PR if that's cool.

PR #300, opened 2025-06-02
Filter buf.gen.yaml files based on protoc version

pkwarren:
Update the buf.gen.yaml files to filter based on the protoc version, removing a manual step required when updating to newer protoc versions.

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -1,5 +1,5 @@
 version: v2
 clean: true
 plugins:
-  - remote: buf.build/protocolbuffers/java:v31.1
+  - remote: buf.build/protocolbuffers/java:$protocJavaPluginVersion
     out: build/generated/sources/bufgen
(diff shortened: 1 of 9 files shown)


buf.gen.yaml
```diff
@@ -1,5 +1,5 @@
 version: v2
 clean: true
 plugins:
-  - remote: buf.build/protocolbuffers/java:v31.1
+  - remote: buf.build/protocolbuffers/java:$protocJavaPluginVersion
```
pkwarren, line 4:
```
$ tree build/buf-gen-templates
build/buf-gen-templates
├── buf.gen.yaml
└── src
    └── test
        └── resources
            └── proto
                ├── buf.gen.cel.testtypes.yaml
                ├── buf.gen.cel.yaml
                ├── buf.gen.imports.yaml
                └── buf.gen.noimports.yaml
$ cat build/buf-gen-templates/buf.gen.yaml 
version: v2
clean: true
plugins:
  - remote: buf.build/protocolbuffers/java:v31.1
    out: build/generated/sources/bufgen
```

PR #303, opened 2025-06-06
Switch to `dev.cel:cel`

srikrsna-buf:
Switch to Google's CEL implementation.

- Type checking is stricter, which meant:
  - Converting to CEL types (Long, UnsignedLong, Double, etc...) before passing it to the runtime. Eg: Floats and Double are different for something as simple as `-` operator and using Floats doesn't work with any function.
  - Changing zero values to CEL supported ones.
  - Declaring separate function overloads based on type. Resulting in a slight increase in function definitions.
- Functions bindings are based on java types.
- Partial evaluation is not well documented so it is not implemented in the PR. Also unsure if it is working on main, more on this in a comment.
- Global variables per program/runtime are not supported. All the variables need to be passed to `eval`.

Deliberately didn't refactor to keep relatively easier to review.

### DIFF
--- a/src/main/java/build/buf/protovalidate/CompiledProgram.java
+++ b/src/main/java/build/buf/protovalidate/CompiledProgram.java
@@ -68,7 +68,7 @@ class CompiledProgram {
   /**
    * Evaluate the compiled program with a given set of {@link Variable} variables.
    *
-   * @param variables Variable variables used for the evaluation.
+   * @param variables Variables used for the evaluation.
    * @param fieldValue Field value to return in violations.
    * @return The {@link build.buf.validate.Violation} from the evaluation, or null if there are no
    *     violations.


conformance/expected-failures.yaml
```diff
@@ -0,0 +1,9 @@
+standard_rules/bytes:
+  - pattern/invalid/not_utf8
+#        input: [ type.googleapis.com/buf.validate.conformance.cases.BytesPattern ]:{val:"\x99"}
+#        want: runtime error: value must be valid UTF-8 to apply regexp
+#        got: validation error (1 violation)
+#         1. rule_id: "bytes.pattern"
+#           message: "value must match regex pattern `^[\\x00-\\x7F]+$`"
+#           field: "val" elements:{field_number:1 field_name:"val" field_type:TYPE_BYTES}
+#           rule: "bytes.pattern" elements:{field_number:15 field_name:"bytes" field_type:TYPE_MESSAGE} elements:{field_number:4 field_name:"pattern" field_type:TYPE_STRING}
```
srikrsna-buf, line 9: Can be solved if we override the `bytes_to_string`: https://github.com/bufbuild/protovalidate-java/pull/304

src/main/java/build/buf/protovalidate/RuleCache.java
```diff
@@ -133,37 +119,19 @@ public List<CompiledProgram> compile(
...
-            continue;
-          }
-        }
-        Ast residual = ruleEnv.residualAst(rule.astExpression.ast, evalResult.getEvalDetails());
-        programs.add(
-            new CompiledProgram(
-                ruleEnv.program(residual, globals),
-                rule.astExpression.source,
-                rule.rulePath,
-                ruleValue));
```
srikrsna-buf, line 159: From an internal feature standpoint this is the only step we skipped. 

I am also not sure if it ever worked. It seems we have to tell which variables/attrs are not available by passing a `PartialActivation` type to `eval`. See [here](https://github.com/projectnessie/cel-java/blob/5e68e2486b3c74490367658ca09e1be82fbb37ab/core/src/main/java/org/projectnessie/cel/EvalOption.java#L42).

src/main/java/build/buf/protovalidate/ProtoAdapter.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * CEL supports protobuf natively but when we pass it field values (like scalars, repeated, and
+ * maps) it has no way to treat them like a proto message field. This class has methods to convert
+ * to a cel values.
+ */
+final class ProtoAdapter {
```
timostamm, line 31: 👍

src/main/java/build/buf/protovalidate/NowVariable.java
```diff
@@ -14,41 +14,39 @@
...
-import org.projectnessie.cel.interpreter.ResolvedValue;
 
 /**
- * {@link NowVariable} implements {@link Activation}, providing a lazily produced timestamp for
- * accessing the variable `now` that's constant within an evaluation.
+ * {@link NowVariable} implements {@link CelVariableResolver}, providing a lazily produced timestamp
+ * for accessing the variable `now` that's constant within an evaluation.
  */
-class NowVariable implements Activation {
+class NowVariable implements CelVariableResolver {
```
timostamm, line 27: IIRC, we had a bug early on where `now` was cached for a validator instance, instead of for a validation pass. The conformance tests can't catch it. Would be good to be sure we don't regress here.

conformance/src/test/java/build/buf/protovalidate/ValidatorTest.java
```diff
@@ -213,10 +213,7 @@ public void strictRepeatedEnumIn() throws Exception {
   @Test
   public void strictRepeatedMin() throws Exception {
     RepeatedMin invalid =
-        RepeatedMin.newBuilder()
-            .addVal(Embed.newBuilder().setVal(1).build())
-            .addVal(Embed.newBuilder().setVal(-1).build())
-            .build();
+        RepeatedMin.newBuilder().addVal(Embed.newBuilder().setVal(1).build()).build();
```
timostamm, line 216: `RepeatedMin` has a `repeated.min_items` rule, and the list element has a `int64.gt` rule. 

I believe the intent of this test is to prove that when the list element rule is violated, `repeated.min_items` doesn't add a violation.

Without going into the details, I think we should simply keep the test as is.

timostamm (commented):
Looks great so far, but I wasn't able to review everything in this pass.

src/main/java/build/buf/protovalidate/NowVariable.java
srikrsna-buf: I did not know that. I checked the references for `NowVariable` class, it is only created [here](https://github.com/bufbuild/protovalidate-java/blob/e13ebd082b058738481a9c9c517d1dd7d0aeb5ee/src/main/java/build/buf/protovalidate/Variable.java#L58) alongside `this` variable.

build.gradle.kts
```diff
@@ -339,12 +339,12 @@ dependencies {
...
     api(libs.jspecify)
     api(libs.protobuf.java)
-    implementation(enforcedPlatform(libs.cel))
-    implementation(libs.cel.core)
+    implementation(libs.cel)
 
     buf("build.buf:buf:${libs.versions.buf.get()}:${osdetector.classifier}@exe")
 
     testImplementation(libs.assertj)
+    testImplementation(libs.grpc.protobuf)
```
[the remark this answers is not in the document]
smaye81: Where is that used?

smaye81 (approved):

jchadwick-buf (approved):
LGTM. I think you could also roll #304 into this too, if you wanted, but fine either way.

build.gradle.kts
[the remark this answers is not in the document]
srikrsna: In some of the cel protos we generate code for

PR #327, opened 2025-07-14
Update to CEL 0.10.1

pkwarren:


srikrsna-buf (commented):
We should be able to remove `celBytesToString` now that https://github.com/google/cel-java/pull/717 has been merged.

pkwarren:
> We should be able to remove `celBytesToString` now that [google/cel-java#717](https://github.com/google/cel-java/pull/717) has been merged.

Fixed in 6c7e45ff

PR #451, opened 2026-04-23
Fix cache key mismatch in RuleCache.compileRule

snuderl:
## Summary

`RuleCache.compileRule()` stores entries in `descriptorMap` keyed by `ruleFieldDesc` (the rule field, e.g. `StringRules.pattern`), but the lookup at the top of the method uses `fieldDescriptor` (the user's field being validated).

The read key and write key never agree, so cached `CelRule` lists are never reused across different user fields that share the same predefined rule — every call falls through and re-builds the AST.

This PR changes the lookup to use `ruleFieldDesc` so it matches the store.

## Test plan

- [x] Existing unit tests pass (`./gradlew test`)
- [x] Existing conformance tests pass

### DIFF
--- a/src/main/java/build/buf/protovalidate/RuleCache.java
+++ b/src/main/java/build/buf/protovalidate/RuleCache.java
@@ -144,7 +144,7 @@ final class RuleCache {
       FieldDescriptor ruleFieldDesc,
       Message message)
       throws CompilationException {
-    List<CelRule> celRules = descriptorMap.get(fieldDescriptor);
+    List<CelRule> celRules = descriptorMap.get(ruleFieldDesc);
     if (celRules != null) {
       return celRules;
     }


pkwarren:
Thank you for the contribution. The fix looks correct to me - I've merged the latest changes from main which resolve an issue with PRs raised from forks although builds may still be rate limited.

Can you sign the CLA?

pkwarren (approved):

pkwarren:
Wiring up a small benchmark harness - verified the fix:
```
benchmark                    metric  before            after            delta
compileValidatorForRepeated  time    4696209.42 ns/op  1064942.2 ns/op  -77.4%
compileValidatorForRepeated  alloc   12950196.95 B/op  3262651.61 B/op  -74.9%
```

PR #452, opened 2026-04-23
Cache compiled programs by serialized rule message

snuderl:
## Summary

Fields with identical resolved rules (same rule type + same values — e.g. multiple fields sharing the same predefined constraint such as an `evm_hash` pattern) currently rebuild their `List<CompiledProgram>` from scratch every time, even though the resulting programs are equivalent.

This PR adds a static `ConcurrentHashMap` in `RuleCache` keyed by the serialized rule message bytes. On hit, `compile()` returns the cached list directly; on miss, it falls through to the existing build path and populates the cache with the result.

In a workload with 14 descriptor types and 19 fields sharing 2 predefined rules, warmup time dropped from ~58ms to ~3.6ms.

Note: this is independent of the cache-key fix in #451 — the two help different scenarios (AST reuse across fields sharing a rule field descriptor vs. program reuse across fields with identical resolved rule values) and can be merged in either order.

## Test plan

- [x] Existing unit tests pass (`./gradlew test`)
- [x] Existing conformance tests pass

### DIFF
--- a/src/main/java/build/buf/protovalidate/RuleCache.java
+++ b/src/main/java/build/buf/protovalidate/RuleCache.java
@@ -64,6 +64,13 @@ final class RuleCache {
   private static final Map<FieldDescriptor, List<CelRule>> descriptorMap =
       new ConcurrentHashMap<>();
 
+  /**
+   * Concurrent map for caching compiled programs by the serialized rule message bytes. This allows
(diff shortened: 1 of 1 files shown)


snuderl:
Superseded by #453, which achieves the same reuse by caching the compiled program in the existing descriptor-keyed cache (matching protovalidate-go's design) instead of adding a second bytes-keyed cache.

PR #454, opened 2026-04-23
Skip tautological evaluators in MessageEvaluator.append

snuderl:
## Summary

`MessageEvaluator.append` stored every evaluator unconditionally, so `FieldEvaluator`s for fields with no rules ran on every `validate()` call: `hasField` check, `ObjectValue` allocation, `ValueEvaluator.evaluate` with a zero-length evaluator list, return empty. None of that can ever produce a violation, so iterating those entries is pure overhead.

`ValueEvaluator.append` already filters on `tautology()`; this mirrors the check in `MessageEvaluator.append` to match protovalidate-go's `AppendNested` ([internal/evaluator/message.go:67-77](https://github.com/bufbuild/protovalidate-go/blob/main/internal/evaluator/message.go#L67-L77)), which skips appends whose `Tautology()` is true.

## Impact

Measured on a message with 2 fields (1 ruled, 1 unruled) over 1M iterations with a warm JIT:
- baseline: ~977 ns/validate
- with filter: ~908 ns/validate (~7%)

The effect scales with the fraction of no-rule fields. Messages with many unconstrained fields (common in real-world proto definitions where validation targets a subset) save proportionally more.

## Test plan

- [x] `./gradlew test` passes (unit + conformance)


pkwarren (approved):
Changes look good to me. Thanks for the fix!

```
> Task :benchmarks:jmhCompare
benchmark                    metric  before            after             delta
validateManyUnruled          time    625.57 ns/op      407.94 ns/op      -34.8%
validateManyUnruled          alloc   3140 B/op         2088 B/op         -33.5%
```

PR #455, opened 2026-04-23
Move BSR login into the buf-action step

pkwarren:
PRs opened from forks currently fail as the BUF_TOKEN secret isn't available. Update the workflows to move BSR login into the buf-action step, which gracefully handles an empty token and skips authentication.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,9 +34,7 @@ jobs:
       - uses: bufbuild/buf-action@fd21066df7214747548607aaa45548ba2b9bc1ff # v1.4.0
         with:
           setup_only: true
-      - env:
-          BUF_TOKEN: ${{ secrets.BUF_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -34,9 +34,7 @@ jobs:
       - uses: bufbuild/buf-action@fd21066df7214747548607aaa45548ba2b9bc1ff # v1.4.0
         with:
           setup_only: true
-      - env:
-          BUF_TOKEN: ${{ secrets.BUF_TOKEN }}
-        run: echo ${BUF_TOKEN} | buf registry login buf.build --token-stdin
+          token: ${{ secrets.BUF_TOKEN }}
```
pkwarren, line 37: https://github.com/bufbuild/buf-action/blob/219c9aef02e8f6a27e4b46274db239ef513f5d53/src/main.ts#L219-L222

stefanvanburen (approved):

PR #456, opened 2026-04-23
Add missing inputs and outputs to Gradle exec tasks

pkwarren:
Update Gradle exec tasks to declare their inputs and outputs, so that tasks won't run multiple times (especially those that interact with the BSR, potentially leading to rate limiting). This is seen in CI runs where `make generate` and `make build` run the same generation and export steps.

### DIFF
--- a/conformance/buf.gen.yaml
+++ b/conformance/buf.gen.yaml
@@ -1,5 +1,4 @@
 version: v2
-clean: true
 managed:
   enabled: true
   override:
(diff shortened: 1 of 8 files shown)


conformance/buf.gen.yaml
```diff
@@ -1,5 +1,4 @@
 version: v2
-clean: true
```
pkwarren, line 2: This isn't needed and wasn't consistently set across other buf.gen.yaml files.

stefanvanburen (approved):

stefanvanburen (approved):

PR #457, opened 2026-04-23
Create a small JMH benchmark harness

pkwarren:
In order to verify performance fixes (like #451 and #454), it would be helpful to be able to write a JMH benchmark and quickly measure the impact.

This adds a benchmark module to the Gradle build. CI only verifies the build of the benchmarks. Added a README showing how to add benchmarks and measure improvements locally.

Instead of outputting JSON or other format, created a jq script to roughly print out output similar to benchstat for Go.


benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,137 @@
+plugins {
+    java
+    alias(libs.plugins.osdetector)
+    id("me.champeau.jmh") version "0.7.2"
```
stefanvanburen: should this be in `gradle/libs.versions.toml` instead?

benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/ValidationBenchmark.java
```diff
@@ -0,0 +1,101 @@
+// Copyright 2023-2026 Buf Technologies, Inc.
```
stefanvanburen, line 1: nit: think we ought to bump [license-header.years](https://github.com/bufbuild/protovalidate-java/blob/9fb2d8cf1d76cd75e0f7ee98f8622e655c944e50/gradle.properties#L8) in general; can be separate?

stefanvanburen (approved):
nothing blocking, just wondering

benchmarks/build.gradle.kts
pkwarren: Good catch - fixed in b1b9dd76.

benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/ValidationBenchmark.java
pkwarren: You're looking at an older version - this is already bumped: https://github.com/bufbuild/protovalidate-java/blob/f05bf0728a6a700524966aa9f07a2a43aec56fa2/gradle.properties#L7-L8

stefanvanburen (approved):

PR #459, opened 2026-04-24
Add a cache for compiled regex patterns

pkwarren:
cel-java doesn't currently have the optimization from cel-go which optimizes calls to the 'matches' function to compile a regular expression once instead of at every invocation.

Override cel-java's builtin matches functions to add a pattern compilation cache (scoped to a Validator).

### DIFF
--- a/benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/ValidationBenchmark.java
+++ b/benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/ValidationBenchmark.java
@@ -17,6 +17,7 @@ package build.buf.protovalidate.benchmarks;
 import build.buf.protovalidate.Validator;
 import build.buf.protovalidate.ValidatorFactory;
 import build.buf.protovalidate.benchmarks.gen.ManyUnruledFieldsMessage;
+import build.buf.protovalidate.benchmarks.gen.RegexPatternMessage;
 import build.buf.protovalidate.benchmarks.gen.RepeatedRuleMessage;
 import build.buf.protovalidate.benchmarks.gen.SimpleStringMessage;
 import build.buf.protovalidate.exceptions.ValidationException;
@@ -40,6 +41,7 @@ public class ValidationBenchmark {
   private SimpleStringMessage simple;
   private ManyUnruledFieldsMessage manyUnruled;
   private RepeatedRuleMessage repeatedRule;
+  private RegexPatternMessage regexPattern;
 
   @Setup
   public void setup() throws ValidationException {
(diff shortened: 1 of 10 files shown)


src/main/java/build/buf/protovalidate/ValidateLibrary.java
```diff
@@ -14,24 +14,59 @@
...
 
 /**
  * Custom {@link CelCompilerLibrary} and {@link CelRuntimeLibrary}. Provides all the custom
  * extension function definitions and overloads.
  */
 final class ValidateLibrary implements CelCompilerLibrary, CelRuntimeLibrary {
 
+  private static final CelOptions CEL_OPTIONS = CelOptions.DEFAULT;
+
+  private final ConcurrentMap<String, Pattern> patternCache = new ConcurrentHashMap<>();
```
pkwarren, line 43: Unlike the proposed fix in https://github.com/google/cel-java/issues/1038, this cache is scoped to a Validator and not global.

benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/ValidationBenchmark.java
```diff
@@ -90,4 +95,9 @@ public void validateManyUnruled(Blackhole bh) throws ValidationException {
   public void validateRepeatedRule(Blackhole bh) throws ValidationException {
     bh.consume(validator.validate(repeatedRule));
   }
+
+  @Benchmark
+  public void validateRegexPattern(Blackhole bh) throws ValidationException {
```
pkwarren, line 100: Results:
```
benchmark             metric  before         after          delta
validateRegexPattern  time    1873.21 ns/op  1323.61 ns/op  -29.3%
validateRegexPattern  alloc   7828 B/op      4272 B/op      -45.4%
```

stolas-dev (approved):

src/test/java/build/buf/protovalidate/CustomOverloadTest.java
```diff
@@ -31,14 +29,7 @@
 
 public class CustomOverloadTest {
 
-  private final ValidateLibrary validateLibrary = new ValidateLibrary();
-  private final Cel cel =
-      CelFactory.standardCelBuilder()
-          .addCompilerLibraries(validateLibrary)
-          .addRuntimeLibraries(validateLibrary)
-          .setOptions(
-              CelOptions.DEFAULT.toBuilder().evaluateCanonicalTypesToNativeValues(true).build())
```
stefanvanburen, line 40: do we still need `evaluateCanonicalTypesToNativeValues(true)` in `newCel`?

rodaine (approved):

src/test/java/build/buf/protovalidate/CustomOverloadTest.java
pkwarren: This is deprecated and the default is true so I went ahead and removed it.

PR #466, opened 2026-04-29
Update to latest licenseheader binary and ignore yaml files that shouldn't have headers

jonbodner-buf:


build.gradle.kts
```diff
@@ -79,6 +79,18 @@ tasks.register<Exec>("licenseHeader") {
...
         "--ignore",
         "src/main/resources/buf/validate/",
+        "--ignore",
+        ".github",
+        "--ignore",
+        "src/test/resources/proto",
+        "--ignore",
+        "benchmarks",
+        "--ignore",
+        "conformance",
```
pkwarren: I don't think we want to ignore all benchmarks and conformance files. Can we be more specific about ignoring yaml files?

jonbodner-buf: good catch; yes.

jonbodner-buf: updated.

pkwarren (approved):

PR #469, opened 2026-05-04
Add native rule implementation for the standard protovalidate rules

jonbodner-buf:
This PR ports the native rule support from protovalidate-go to protovalidate-java. Validation rules defined by protovalidate are handled in Java code instead of CEL. CEL expressions are still processed in CEL, as are any new rules that don't have native implementations.

The standard rules can be evaluated either through CEL or through native Java code. Native evaluation is functionally identical (the conformance suite passes in both modes) but skips CEL compilation and runtime overhead for the rules it covers — a single validate() call on a complex message can run an order of magnitude faster and allocate ~10× less.

Native rules are opt-in while the implementation matures. Enable them by configuring the validator:

```
Config config = Config.newBuilder().setEnableNativeRules(true).build();
Validator validator = ValidatorFactory.newBuilder().withConfig(config).build();
```

Native evaluators currently cover:
- bool
- all 12 numeric kinds (signed and unsigned int32/int64, float, double, etc.)
- enum (const/in/not_in; the existing defined_only path is unchanged)
- bytes
- string (including all well-known formats)
- repeated/map list-level rules (min_items/max_items/unique, min_pairs/max_pairs)
-  the google.protobuf.{Bool,Int32,Int64,UInt32,UInt64,Float,Double,String,Bytes}Value wrapper types (rules on wrapper-typed fields run through the same native evaluators after unwrapping).

Forward compatibility is preserved by a clone-and-clear contract: when protovalidate adds a new rule that this codebase hasn't yet implemented natively, the rule remains on the residual FieldRules and CEL enforces it. Native evaluation is an optimization, never a replacement.

Performance improvement for native rule implementations:

```
benchmark                          metric  cel                native          delta
buildBenchComplexSchema            time    15002410.58 ns/op  36947.91 ns/op  -99.8%
buildBenchComplexSchema            alloc   22874194.63 B/op   74972.01 B/op   -99.7%
buildBenchInt32GT                  time    4978570.44 ns/op   15568.07 ns/op  -99.7%
buildBenchInt32GT                  alloc   8468429.34 B/op    43944 B/op      -99.5%
validateBenchBoolConst             time    506.87 ns/op       60.67 ns/op     -88%
validateBenchBoolConst             alloc   1552 B/op          8 B/op          -99.5%
validateBenchBytesConst            time    614.72 ns/op       96.61 ns/op     -84.3%
validateBenchBytesConst            alloc   1828 B/op          140 B/op        -92.3%
validateBenchBytesIn               time    1498.31 ns/op      126.79 ns/op    -91.5%
validateBenchBytesIn               alloc   4288 B/op          168 B/op        -96.1%
validateBenchComplexSchema         time    54851.83 ns/op     4835.59 ns/op   -91.2%
validateBenchComplexSchema         alloc   130384.01 B/op     9184 B/op       -93%
validateBenchDoubleIn              time    686.87 ns/op       84 ns/op        -87.8%
validateBenchDoubleIn              alloc   2324 B/op          140 B/op        -94%
validateBenchEnumConst             time    506.34 ns/op       64.54 ns/op     -87.3%
validateBenchEnumConst             alloc   1556 B/op          72 B/op         -95.4%
validateBenchEnumNotIn             time    505.09 ns/op       65.28 ns/op     -87.1%
validateBenchEnumNotIn             alloc   1620 B/op          80 B/op         -95.1%
validateBenchEnumRules             time    726.41 ns/op       67.74 ns/op     -90.7%
validateBenchEnumRules             alloc   1960 B/op          80 B/op         -95.9%
validateBenchInt32GT               time    36965.57 ns/op     1114.67 ns/op   -97%
validateBenchInt32GT               alloc   85736.01 B/op      1244 B/op       -98.5%
validateBenchInt64Const            time    556.7 ns/op        77.81 ns/op     -86%
validateBenchInt64Const            alloc   1740 B/op          72 B/op         -95.9%
validateBenchInt64In               time    708.46 ns/op       94.13 ns/op     -86.7%
validateBenchInt64In               alloc   2048 B/op          132 B/op        -93.6%
validateBenchMap                   time    1007.53 ns/op      226.63 ns/op    -77.5%
validateBenchMap                   alloc   2912 B/op          432 B/op        -85.2%
validateBenchRepeatedBytesUnique   time    1034.72 ns/op      326.8 ns/op     -68.4%
validateBenchRepeatedBytesUnique   alloc   3304 B/op          1088 B/op       -67.1%
validateBenchRepeatedInt32Unique   time    668.98 ns/op       205.97 ns/op    -69.2%
validateBenchRepeatedInt32Unique   alloc   2256 B/op          576 B/op        -74.5%
validateBenchRepeatedMessage       time    9555.54 ns/op      645.4 ns/op     -93.2%
validateBenchRepeatedMessage       alloc   31112 B/op         1012 B/op       -96.7%
validateBenchRepeatedScalar        time    688.25 ns/op       95.64 ns/op     -86.1%
validateBenchRepeatedScalar        alloc   2240 B/op          112 B/op        -95%
validateBenchRepeatedScalarUnique  time    685.01 ns/op       203.59 ns/op    -70.3%
validateBenchRepeatedScalarUnique  alloc   2592 B/op          856 B/op        -67%
validateBenchRepeatedStringUnique  time    637.11 ns/op       232.19 ns/op    -63.6%
validateBenchRepeatedStringUnique  alloc   2176 B/op          576 B/op        -73.5%
validateBenchScalar                time    1523.08 ns/op      85.91 ns/op     -94.4%
validateBenchScalar                alloc   4492 B/op          120 B/op        -97.3%
validateBenchStringConst           time    548.64 ns/op       78.3 ns/op      -85.7%
validateBenchStringConst           alloc   1728 B/op          72 B/op         -95.8%
validateBenchStringContains        time    498.97 ns/op       74.8 ns/op      -85%
validateBenchStringContains        alloc   1628 B/op          72 B/op         -95.6%
validateBenchStringIn              time    778.2 ns/op        80.58 ns/op     -89.6%
validateBenchStringIn              alloc   2176 B/op          100 B/op        -95.4%
validateBenchStringLen             time    619.07 ns/op       70.94 ns/op     -88.5%
validateBenchStringLen             alloc   2132 B/op          72 B/op         -96.6%
validateBenchStringMinLen          time    640.65 ns/op       69.71 ns/op     -89.1%
validateBenchStringMinLen          alloc   2236 B/op          72 B/op         -96.8%
validateBenchStringPrefix          time    517.29 ns/op       79.1 ns/op      -84.7%
validateBenchStringPrefix          alloc   1732 B/op          72 B/op         -95.8%
validateBenchUint32In              time    747.06 ns/op       86.95 ns/op     -88.4%
validateBenchUint32In              alloc   2120 B/op          124 B/op        -94.2%
validateManyUnruled                time    650.71 ns/op       68.37 ns/op     -89.5%
validateManyUnruled                alloc   2192 B/op          72 B/op         -96.7%
validateMultiRuleError             time    3292.56 ns/op      433.89 ns/op    -86.8%
validateMultiRuleError             alloc   10736 B/op         2168 B/op       -79.8%
validateMultiRuleNoError           time    1958.92 ns/op      82.84 ns/op     -95.8%
validateMultiRuleNoError           alloc   5872 B/op          72 B/op         -98.8%
validateRegexPattern               time    1987 ns/op         779.5 ns/op     -60.8%
validateRegexPattern               alloc   4296 B/op          180 B/op        -95.8%
validateRepeatedRule               time    14083.24 ns/op     1186.26 ns/op   -91.6%
validateRepeatedRule               alloc   42896 B/op         1468 B/op       -96.6%
validateSimple                     time    2130.52 ns/op      882.01 ns/op    -58.6%
validateSimple                     alloc   4056 B/op          152 B/op        -96.3%
validateStringMatching             time    8325.98 ns/op      2035.29 ns/op   -75.6%
validateStringMatching             alloc   16280 B/op         960 B/op        -94.1%
validateTestByteMatching           time    6546.92 ns/op      359.91 ns/op    -94.5%
validateTestByteMatching           alloc   22120 B/op         600 B/op        -97.3%
validateWrapperTesting             time    18270 ns/op        962.24 ns/op    -94.7%
validateWrapperTesting             alloc   40404 B/op         1376 B/op       -96.6%
```

Run the benchmarks yourself with:

```
./gradlew :benchmarks:jmh
./gradlew :benchmarks:jmhCompareParams
```


jonbodner-buf:
New commit reverses the behavior. Now native rules are enabled by default.

.github/workflows/conformance.yaml
```diff
@@ -37,5 +37,7 @@ jobs:
           token: ${{ secrets.BUF_TOKEN }}
       - name: Validate Gradle Wrapper
         uses: gradle/actions/wrapper-validation@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6.1.0
-      - name: Test conformance
+      - name: Test conformance (CEL-only)
+        run: make conformance-cel
+      - name: Test conformance (native rules enabled — default mode)
```
pkwarren: nit: I'd remove the `(native rules enabled ...` here - there will potentially be other variants down the road.

Makefile
```diff
@@ -28,9 +28,13 @@ checkgenerate: generate  ## Checks if `make generate` produces a diff.
...
 
+.PHONY: conformance-cel
+conformance-cel: ## Execute conformance tests with CEL-only rule evaluation.
+	ENABLE_NATIVE_RULES=false $(GRADLE) conformance:conformance
+
 .PHONY: conformance
-conformance: ## Execute conformance tests.
-	$(GRADLE) conformance:conformance
+conformance: ## Execute conformance tests with native rule evaluators enabled.
+	ENABLE_NATIVE_RULES=true $(GRADLE) conformance:conformance
```
pkwarren, line 33: We could define `ENABLE_NATIVE_RULES ?= true` above and then just update the workflow to run `make conformance ENABLE_NATIVE_RULES=false` for the CEL flow.

README.md
```diff
@@ -64,6 +64,21 @@ Highlights for Java developers include:
 * A comprehensive RPC quickstart for [Java and gRPC][grpc-java]
 * A [migration guide for protoc-gen-validate][migration-guide] users
 
+## Native rule evaluators (opt-in)
```
pkwarren: Now opt-out

README.md
```diff
@@ -64,6 +64,21 @@ Highlights for Java developers include:
...
+The standard rules can be evaluated either through CEL or through native Java code. Native evaluation is functionally identical (the conformance suite passes in both modes) but skips CEL compilation and runtime overhead for the rules it covers — a single `validate()` call on a complex message can run an order of magnitude faster and allocate ~10× less.
+
+Native rules are **opt-out**. Disable them by configuring the validator:
+
+```java
+Config config = Config.newBuilder().setDisableNativeRules().build();
+Validator validator = ValidatorFactory.newBuilder().withConfig(config).build();
+```
+
+Native evaluators currently cover bool, all 12 numeric kinds (signed and unsigned int32/int64, float, double, etc.), enum (`const`/`in`/`not_in`; the existing `defined_only` path is unchanged), bytes, string (including all well-known formats), repeated/map list-level rules (`min_items`/`max_items`/`unique`, `min_pairs`/`max_pairs`), and the `google.protobuf.{Bool,Int32,Int64,UInt32,UInt64,Float,Double,String,Bytes}Value` wrapper types — rules on wrapper-typed fields run through the same native evaluators after unwrapping.
```
pkwarren: Is this helpful information for users or a detail left to the implementation's javadoc or somewhere else?

My main concern is keeping this list in sync with any future optimizations.

benchmarks/build.gradle.kts
```diff
@@ -143,3 +143,26 @@ tasks.register<Exec>("jmhCompare") {
...
 }
+
+// Diffs the two enableNativeRules variants within a single results.json.
+// `before` column is CEL (enableNativeRules=false), `after` is native
+// (enableNativeRules=true), so a negative delta means native is faster /
+// allocates less.
+//
+// Override the input file:
+//   ./gradlew :benchmarks:jmhCompareParams -Presults=path/to/results.json
+tasks.register<Exec>("jmhCompareParams") {
```
pkwarren: Should we call this `jmhCompareNative` or `jmhCompareNativeRules`?

If so we should update the naming of the jq script to match.

benchmarks/jmh-compare-native-rules.jq
```diff
@@ -0,0 +1,27 @@
+def pct(a; b):
```
pkwarren, line 6: It might be good to add a short comment above on how this compares with the other jq script.

conformance/src/main/java/build/buf/protovalidate/conformance/Main.java
```diff
@@ -61,11 +61,21 @@ static TestConformanceResponse testConformance(TestConformanceRequest request) {
...
       ExtensionRegistry extensionRegistry =
           FileDescriptorUtil.createExtensionRegistry(fileDescriptorMap.values());
-      Config cfg =
-          Config.newBuilder()
-              .setTypeRegistry(typeRegistry)
-              .setExtensionRegistry(extensionRegistry)
-              .build();
+      // ENABLE_NATIVE_RULES env var lets the conformance runner exercise both rule-evaluation
+      // modes without code changes. Defaults to whatever Config's default is so a plain
+      // `gradlew :conformance:test` matches user-facing behavior.
```
pkwarren: nit: This feels a little over commented - it is pretty easy to tell what it is doing and not sure if we need the comment.

conformance/src/test/java/build/buf/protovalidate/NativeRulesParityTest.java
```diff
@@ -0,0 +1,139 @@
...
+  @MethodSource("fixtures")
+  void parityForFixture(String name, Message msg) throws ValidationException {
+    ValidationResult nativeResult = nativeValidator.validate(msg);
+    ValidationResult celResult = celValidator.validate(msg);
+    assertThat(toProtoList(nativeResult))
+        .as("toProto() parity for %s", name)
+        .isEqualTo(toProtoList(celResult));
+  }
+
+  private static List<build.buf.validate.Violation> toProtoList(ValidationResult result) {
```
pkwarren, line 136: Can we used the unqualified `Violation` here?

src/main/java/build/buf/protovalidate/BytesRulesEvaluator.java
```diff
@@ -0,0 +1,518 @@
...
+  private final RuleBase base;
+  private final @Nullable ByteString constVal;
+  private final @Nullable Long exactLen;
+  private final @Nullable Long minLen;
+  private final @Nullable Long maxLen;
+  private final @Nullable Pattern pattern;
+  private final @Nullable String patternStr;
+  private final @Nullable ByteString prefix;
+  private final @Nullable ByteString suffix;
+  private final @Nullable ByteString contains;
```
pkwarren, line 163: If this class is created for every bytes field that uses validation, these could add up in size. I wonder if there's another pattern we could try to only create data for the present rules.

For example, if we used a separate BitSet for the present rules, then we could change all the `Long` to `long` to avoid boxing.

Just a thought - don't need to change anything right now.

src/main/java/build/buf/protovalidate/EnumRulesEvaluator.java
```diff
@@ -0,0 +1,170 @@
...
+    if (constVal != null && actual != constVal) {
+      RuleViolation.Builder b =
+          NativeViolations.newViolation(CONST_SITE, null, "must equal " + constVal, val, constVal);
+      violations = RuleBase.add(violations, b);
+      if (failFast) {
+        return base.done(violations);
+      }
+    }
+
+    if (!inVals.isEmpty() && !inVals.contains(actual)) {
```
pkwarren, line 122: Probably unnecessary but if we switched from `List` -> `Collection` as the field type, then if there are a large number of in/out values, we could switch at rule creation to to a Set.

Another option might be to avoid boxing and switch to `int[]` and use `Arrays.binarySearch` here. All based on benchmarking of course.

src/main/java/build/buf/protovalidate/EnumRulesEvaluator.java
```diff
@@ -0,0 +1,170 @@
...
+   */
+  private static int enumNumber(Object raw) {
+    if (raw instanceof EnumValueDescriptor) {
+      return ((EnumValueDescriptor) raw).getNumber();
+    }
+    if (raw instanceof Integer) {
+      return (Integer) raw;
+    }
+    if (raw instanceof Long) {
+      return ((Long) raw).intValue();
```
pkwarren: Not sure why Long would be used here but we probably want to do `Math.toIntExact` and rethrow any exceptions instead of just casting here.

src/main/java/build/buf/protovalidate/NumericRulesEvaluator.java
```diff
@@ -0,0 +1,410 @@
...
+    }
+    boolean failure;
+    if (isNormalRange()) {
+      failure = isNaN || aboveHi(actual) || belowLo(actual);
+    } else {
+      failure = isNaN || (aboveHi(actual) && belowLo(actual));
+    }
+    if (failure) {
+      String message = "must be " + loMessage() + " " + conjunction() + " " + hiMessage();
+      return NativeViolations.newViolation(loSite(), gtltRule(), message, val, loVal);
```
pkwarren, line 316: Is `loSite` always the right location? If the value is above the low range then shouldn't this point to the `hiSite`?

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
```diff
@@ -0,0 +1,293 @@
...
+  private static final int FLOAT_NEG_ZERO_BITS = Float.floatToIntBits(-0.0f);
+  private static final long DOUBLE_NEG_ZERO_BITS = Double.doubleToLongBits(-0.0);
+
+  private static String floatFormatter(Float f) {
+    if (Float.floatToIntBits(f) == FLOAT_NEG_ZERO_BITS) {
+      return "-0";
+    }
+    // Whole-number short-circuit: print "5" rather than "5.0" to match Go's %g behavior.
+    float asInt = f.intValue();
+    if (asInt == f) {
```
pkwarren: I think there are edge cases where this will return the wrong value - `2147483648.0f` is one example. We need to check if the values are in range of Integer.MIN/MAX value here.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
```diff
@@ -0,0 +1,293 @@
...
+      return String.valueOf(f.intValue());
+    }
+    return String.valueOf(f);
+  }
+
+  private static String doubleFormatter(Double d) {
+    if (Double.doubleToLongBits(d) == DOUBLE_NEG_ZERO_BITS) {
+      return "-0";
+    }
+    double asInt = d.intValue();
```
pkwarren: Should this be .longValue()?

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  private static final int FLOAT_NEG_ZERO_BITS = Float.floatToIntBits(-0.0f);
+  private static final long DOUBLE_NEG_ZERO_BITS = Double.doubleToLongBits(-0.0);
+
+  private static String floatFormatter(Float f) {
+    if (Float.floatToIntBits(f) == FLOAT_NEG_ZERO_BITS) {
+      return "-0";
+    }
+    // Whole-number short-circuit: print "5" rather than "5.0" to match Go's %g behavior.
```
pkwarren: This may deviate some from Go's behavior - we should check `%g` output against larger values (which are formatted as exponents), +/ Infinity, etc.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
```diff
@@ -0,0 +1,293 @@
...
+          Double.class,
+          NumericTypeConfig::doubleCompare,
+          NumericTypeConfig::doubleFormatter,
+          true);
+
+  // Float and double comparators treat +0.0 and -0.0 as equal, matching IEEE-754. NaN keeps
+  // Java's compareTo semantics (NaN.compareTo(NaN) == 0) so behavior matches the existing
+  // protovalidate-java CEL path, which uses the same Object.equals semantics.
+
+  private static int floatCompare(Float f1, Float f2) {
```
pkwarren, line 255: Since the comparator is used for more than range operations, we may need special handling for NaN in places where the comparator is used.

`Double.NaN.compareTo(Double.NaN) == 0` in Java but if the Go code is using `==` then we'll get different behavior.

src/main/java/build/buf/protovalidate/Config.java
```diff
@@ -157,13 +175,35 @@ public Builder setAllowUnknownFields(boolean allowUnknownFields) {
       return this;
     }
 
+    /**
+     * Enables native (non-CEL) rule evaluators. Forward-compatible: any rule not yet implemented
+     * natively continues to be enforced via CEL regardless of this setting.
+     *
+     * @return this builder
+     */
+    public Builder setEnableNativeRules() {
```
pkwarren: Having two methods here is a different pattern than `setAllowUnknownFields` and `setFailFast` which take the boolean value. I'd have this match and remove `setDisableNativeRules`.

src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
```diff
@@ -71,6 +73,7 @@ final class EvaluatorBuilder {
   EvaluatorBuilder(Cel cel, Config config) {
```
pkwarren, line 72: Not from your change but we should change this to call `this(cel, config, Collections.emptyList(), false)` so we just do this initialization in one place.

src/main/java/build/buf/protovalidate/Internal.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Marks types and members that are public for the protovalidate-java codebase's own use (e.g.
+ * cross-package references between {@code build.buf.protovalidate} and its sub-packages) but are
+ * not part of the supported public API. Code outside this library must not depend on these — they
+ * may change or disappear at any time without notice.
+ */
+@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.FIELD})
+public @interface Internal {}
```
pkwarren: While it makes the package bigger I wonder if we could look to move the rules.* package into build.buf.protovalidate. This way we can continue to leave some classes/methods package private.

There are also some other things we could look at as well - https://errorprone.info/api/latest/com/google/errorprone/annotations/RestrictedApi.html.

src/main/java/build/buf/protovalidate/StringRulesEvaluator.java
```diff
@@ -0,0 +1,836 @@
...
+  // --- Static regexes (compile once) ---
+
+  private static final Pattern UUID_REGEX =
+      Pattern.compile(
+          "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
+  private static final Pattern TUUID_REGEX = Pattern.compile("^[0-9a-fA-F]{32}$");
+  private static final Pattern ULID_REGEX =
+      Pattern.compile("^[0-7][0-9A-HJKMNP-TV-Za-hjkmnp-tv-z]{25}$");
+  private static final Pattern HEADER_NAME_REGEX =
+      Pattern.compile("^:?[0-9a-zA-Z!#$%&\\\\'*+\\-.\\^_|~`]+$");
```
pkwarren: https://github.com/bufbuild/protovalidate-go/blob/c0c9f624fd0fc37177c680ec2e40717ecbc68ac4/native_string.go#L246 uses raw strings, so it appears the escaping of the hyphen char differs betweeen Go and Java.

pkwarren (commented):
Took a first pass - still need to review the tests. The results look really promising - nice work!

There were a few places that may differ a bit between Java and Go that need a closer look. I also think we may be able to optimize further but that's better in follow ups.

src/main/java/build/buf/protovalidate/Config.java
jonbodner-buf: updated. Now there is a single method, `setEnableNativeRules` that takes a `boolean`. The default value is `true`.

conformance/src/test/java/build/buf/protovalidate/NativeRulesParityTest.java
jonbodner-buf: there's a collision with the `Violation` interface in the package, which is referred to in the `map` call below (`Violation::toProto`). Let me see what I can do.

jonbodner-buf: The Violation interface has a single implementation. Not sure it carries its weight, but's used in a few places.

.github/workflows/conformance.yaml
jonbodner-buf: done

benchmarks/build.gradle.kts
jonbodner-buf: renamed to jmhCompareNativeRules, .jq file renamed to jmh-compare-native-rules.jq

benchmarks/jmh-compare-native-rules.jq
jonbodner-buf: comment added

conformance/src/main/java/build/buf/protovalidate/conformance/Main.java
jonbodner-buf: comment removed

src/main/java/build/buf/protovalidate/BytesRulesEvaluator.java
jonbodner-buf: yeah, let's do an optimization pass after the first version lands.

src/main/java/build/buf/protovalidate/EnumRulesEvaluator.java
jonbodner-buf: This is a case where I don't know if Java protobufs could ever have a long value. Added a comment and switched to Math.toIntExact.

src/main/java/build/buf/protovalidate/EvaluatorBuilder.java
jonbodner-buf: added a new private constructor that is called by the two existing constructors. One constructor throws a checked exception, so merging them causes a great deal of disruption as the exception is bubbled up.

src/main/java/build/buf/protovalidate/NumericRulesEvaluator.java
jonbodner-buf: I double-checked and it matches the logic in the Go code. Now checking my test cases to see if they cover all possible cases...

jonbodner-buf: looking back at validate.proto, I'm pretty sure this implementation is correct. hiSite is only used for cases where there is ONLY lt/lte. If there is gt/gte only or a combination of gt/gte and lt/lte, then you are in the loSite. For example:

```
      (predefined).cel = {
        id: "float.gt_lt_exclusive"
        expression:
          "has(rules.lt) && rules.lt < rules.gt && (this.isNan() || (rules.lt <= this && this <= rules.gt))"
          "? 'must be greater than %s or less than %s'.format([rules.gt, rules.lt]) : ''"
      },
```

It doesn't matter whether it is lower or higher, it's the same message and the same id.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: I just double-checked and CEL does the right thing with NaN. I have updated the Java code to also do the right thing.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: Good point, float will fail with some numbers > 2^23 or 24 (forget which), and double will fail after 2^53. I'll add some tests to check and fix the behavior, see if %g in Java matches %g in Go.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: ok, I think this should be working correctly now.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: I think this will be working correctly now.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: yes, but code has been replaced, so not an issue

src/main/java/build/buf/protovalidate/StringRulesEvaluator.java
jonbodner-buf: Let me check if there's a test to validate this.

Makefile
jonbodner-buf: The other option is doing what I did in Go, which is run both conformance tests all the time. what do you think?

README.md
jonbodner-buf: fixed

benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/EvaluatorBuildBenchmark.java
```diff
@@ -52,9 +52,9 @@ public class EvaluatorBuildBenchmark {
   @Setup
   public void setup() {
     if (enableNativeRules) {
-      config = Config.newBuilder().setEnableNativeRules().build();
+      config = Config.newBuilder().setEnableNativeRules(true).build();
     } else {
-      config = Config.newBuilder().setDisableNativeRules().build();
+      config = Config.newBuilder().setEnableNativeRules(false).build();
```
pkwarren: Shouldn't we simplify to do: `config = Config.newBuilder().setEnableNativeRules(enableNativeRules).build()`?

benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/ValidationBenchmark.java
```diff
@@ -131,9 +131,9 @@ public class ValidationBenchmark {
   public void setup() throws ValidationException {
     Config config;
     if (enableNativeRules) {
-      config = Config.newBuilder().setEnableNativeRules().build();
+      config = Config.newBuilder().setEnableNativeRules(true).build();
```
pkwarren: Same here - can simplify to remove if/else.

conformance/src/main/java/build/buf/protovalidate/conformance/Main.java
```diff
@@ -61,18 +61,15 @@ static TestConformanceResponse testConformance(TestConformanceRequest request) {
...
-      // modes without code changes. Defaults to whatever Config's default is so a plain
-      // `gradlew :conformance:test` matches user-facing behavior.
       String envFlag = System.getenv("ENABLE_NATIVE_RULES");
       Config.Builder cfgBuilder =
           Config.newBuilder().setTypeRegistry(typeRegistry).setExtensionRegistry(extensionRegistry);
       if (envFlag != null) {
         boolean enableNativeRules = Boolean.parseBoolean(envFlag);
         if (enableNativeRules) {
-          cfgBuilder.setEnableNativeRules();
+          cfgBuilder.setEnableNativeRules(true);
```
pkwarren: Same here - we can remove the if/else.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
```diff
@@ -255,13 +256,22 @@ private static int floatCompare(Float f1, Float f2) {
...
 
   private static int doubleCompare(Double d1, Double d2) {
     if (d1 == 0.0 && d2 == 0.0) {
       return 0;
     }
+    // NaN != Nan, but Java thinks it does.
+    if (d1.isNaN() && d2.isNaN()) {
+      return -1;
+    }
+
```
pkwarren: nit: Can remove this extra space.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
```diff
@@ -255,13 +256,22 @@ private static int floatCompare(Float f1, Float f2) {
     if (f1 == 0.0f && f2 == 0.0f) {
       return 0;
     }
+    // NaN != Nan, but Java thinks it does.
```
pkwarren, line 259: To be fair, `Float.NaN != Float.NaN`. This is documented as behavior from `compareTo`: https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#compareTo-java.lang.Float-.

src/main/java/build/buf/protovalidate/ValueEvaluator.java
```diff
@@ -25,8 +25,12 @@
 /**
  * {@link ValueEvaluator} performs validation on any concrete value contained within a singular
  * field, repeated elements, or the keys/values of a map.
+ *
+ * <p>Public so that native rule evaluators in {@code build.buf.protovalidate.rules} can be
```
ajeetdsouza: We don't seem to have a separate `rules` package - we can probably make this `package-private` again.

There are a number of such changes throughout the PR - can we fix those?

src/main/java/build/buf/protovalidate/ValidationResult.java
```diff
@@ -26,9 +26,10 @@
 public class ValidationResult {
 
   /**
-   * violations is a list of {@link Violation} that occurred during the validations of a message.
+   * violations is a list of {@link RuleViolation} that occurred during the validations of a
+   * message.
    */
-  private final List<Violation> violations;
+  private final List<RuleViolation> violations;
```
pkwarren: I think we should revert this change. This is part of the public contract and `RuleViolation` is package private.

src/main/java/build/buf/protovalidate/ValueEvaluator.java
```diff
@@ -30,7 +30,6 @@
  * constructed with descriptor/nested-rule context from a {@link ValueEvaluator}; not part of the
  * supported public API.
  */
-@Internal
 public final class ValueEvaluator implements Evaluator {
```
pkwarren: Can we move this back to package private and remove the javadoc comments if they're not needed.

Makefile
pkwarren: That sounds ok. It adds a little to the runtime locally but I can't think when we wouldn't want both to be passing.

src/main/java/build/buf/protovalidate/ValueEvaluator.java
jonbodner-buf: done

src/main/java/build/buf/protovalidate/ValidationResult.java
jonbodner-buf: done

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: IIRC, NaN is a mess in Java, where primitives and wrappers behave differently and different methods of comparison behave differently. I'm OK with the current code, if you are.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: done

conformance/src/main/java/build/buf/protovalidate/conformance/Main.java
jonbodner-buf: done

benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/ValidationBenchmark.java
jonbodner-buf: done

benchmarks/src/jmh/java/build/buf/protovalidate/benchmarks/EvaluatorBuildBenchmark.java
jonbodner-buf: done

pkwarren (approved):

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
```diff
@@ -255,13 +256,22 @@ private static int floatCompare(Float f1, Float f2) {
...
+    if (f1.isNaN() && f2.isNaN()) {
```
ajeetdsouza, line 260: Do we have a conformance test for this? The behavior here should match the behavior in the CEL implementation. (going by the commit message `Native is now more correct than CEL`).

README.md
ajeetdsouza: Why make this opt-in?

src/test/resources/proto/validationtest/validationtest.proto
```diff
@@ -134,3 +135,247 @@ message ExampleImportMessageInMapFieldRule {
     }
   ];
 }
+
+message ExampleBoolConst {
```
ajeetdsouza, line 139: Would it be possible to move some of these tests to the upstream Protovalidate conformance test suite?

README.md
jonbodner-buf: It's opt-out now. My impulse was to make it opt-in in case there's a bug that wasn't found by the conformance suite.

src/test/resources/proto/validationtest/validationtest.proto
jonbodner-buf: I'm working on improving the conformance suite but the work isn't quite ready yet.

src/main/java/build/buf/protovalidate/NumericTypeConfig.java
jonbodner-buf: Not a public one yet.

src/main/java/build/buf/protovalidate/ValueEvaluator.java
jonbodner-buf: This is outdated; the rules package has been merged into the protovalidate package to keep the API from changing.

src/main/java/build/buf/protovalidate/StringRulesEvaluator.java
jonbodner-buf: There wasn't a test. Now there is. There was a problem with the regex. The regex has been fixed and there's an extensive test suite now.

pkwarren (approved):

pkwarren (approved):

PR #480, opened 2026-06-14
Introduce ValidateMessage to enable alternative protobuf runtimes

andrewparmet:
Adds a runtime-independent `ValidateMessage` interface so alternative protobuf runtimes (for example, protokt) can be validated without first converting their messages to `com.google.protobuf.Message`.

## Changes

- Add the public `ValidateMessage` and `Value` surface: `getDescriptorForType`, `hasField`, `getField`, `celValue`, plus `rawValue` / `jvmValue` for the value side.
- Add `ProtobufValidateMessage`, which adapts `com.google.protobuf.Message` to `ValidateMessage`, so the existing `Message` validation path is unchanged.
- Route the evaluators (`FieldEvaluator`, `OneofEvaluator`, `MessageOneofEvaluator`, `AnyEvaluator`, `MapEvaluator`, list/object values, `WrappedValueEvaluator`) through `ValidateMessage` instead of `Message` directly.
- Expose `Validator.validate(ValidateMessage)` as the entry point for alternative-runtime callers; the `Message` overload delegates through it.

## Related work

- protovalidate-java: #481 moves the CEL setup to the planner runtime; the planner is what protokt's `ValidateMessage` implementation runs against.
- protokt: open-toast/protokt#402 is the downstream consumer that validates protokt messages through `ValidateMessage`.
- Supersedes #202, an earlier take on the same `ValidateMessage` interface.

## Testing

`./gradlew :conformance:conformance` passes.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
           setup_only: true
           token: ${{ secrets.BUF_TOKEN }}
       - name: Validate Gradle Wrapper
-        uses: gradle/actions/wrapper-validation@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6.1.0
+        uses: gradle/actions/wrapper-validation@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
       - name: Lint
         run: make lint
       - name: Generate
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -36,6 +36,6 @@ jobs:
           setup_only: true
           token: ${{ secrets.BUF_TOKEN }}
       - name: Validate Gradle Wrapper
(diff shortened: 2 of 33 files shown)


jonbodner-buf:
@andrewparmet can you update the branch from main?

andrewparmet:
@jonbodner-buf done!

jonbodner-buf:
My biggest concern is there are no tests to validate these changes. Yes, the existing tests still pass, but nothing in the existing tests calls `public ValidationResult validate(MessageReflector message) throws ValidationException` . 

The exception thrown by `build.buf.protovalidate.MessageValue#jvmValue` and `build.buf.protovalidate.MessageReflectorValue#jvmValue` also concern me, want to make sure that there is no code path where that could be triggered, especially for MessageReflectorValue, since it will be used with external types that implement `MessageReflector`.

andrewparmet:
Pushed up some tests - thanks for the starter patch from your investigation! The existing Validator tests do indirectly call `validate(ValidateMessage message)` via the Message overload, which hands off directly to it.

PR #481, opened 2026-06-14
Use the CEL planner runtime

andrewparmet:
Moves protovalidate-java's CEL setup from the deprecated standard runtime to the planner runtime.

## Notable callouts

- Wire the compiler and runtime by hand rather than through `plannerCelBuilder()`. `CelRuntimeImpl` directs callers to subset the standard library via `setStandardFunctions` rather than `setStandardEnvironmentEnabled`, so the runtime drops stdlib `matches` that way while the checker continues to use `setStandardEnvironmentEnabled(false)`. This keeps `CustomOverload`'s caching `matches` replacement (cel-java#1038).
- Set `CelOptions.current().enableHeterogeneousNumericComparisons(true)`, which `CelRuntimeImpl` requires.
- Update `cel-java` to 0.13.1 preparing to support alternative Protobuf runtimes.

## Related work:

- cel-java: 0.13.1 carries the CelValue-path fixes this work relies on (fixed32/64 treated as unsigned; FieldMask navigable as a message, google/cel-java#1074).
- protovalidate-java: #480 introduces `MessageReflector` to support alternative Protobuf runtimes, which builds on this runtime change (supersedes #202).
- protokt: open-toast/protokt#402 is the downstream consumer that validates protokt messages through `MessageReflector` on this planner runtime.

## Testing

`./gradlew :conformance:conformance` passes.

### DIFF
--- a/src/main/java/build/buf/protovalidate/ValidateLibrary.java
+++ b/src/main/java/build/buf/protovalidate/ValidateLibrary.java
@@ -18,15 +18,21 @@ import com.google.re2j.Pattern;
 import dev.cel.bundle.Cel;
 import dev.cel.bundle.CelFactory;
 import dev.cel.checker.CelCheckerBuilder;
+import dev.cel.checker.CelCheckerLegacyImpl;
 import dev.cel.checker.CelStandardDeclarations;
 import dev.cel.common.CelOptions;
 import dev.cel.common.CelVarDecl;
 import dev.cel.common.types.SimpleType;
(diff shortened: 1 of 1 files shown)


pkwarren:
@andrewparmet I updated CEL in #482 so this change no longer cleanly applies. Do you mind fixing the conflicts and I'll take a look?

andrewparmet:
@pkwarren done!

pkwarren (approved):

